PR #108, opened 2023-06-29
Trigger Client

frankfka:


packages/trigger-client/src/TriggerClient.ts
```diff
@@ -0,0 +1,283 @@
...
+
+    return {
+      orders,
+    };
+  }
+
+  /*
+  Base Fns
+   */
+  protected async sign<T extends SignableRequestType>(
```
yruej301, line 191: could we extract these in a base class?

yruej301 (approved):
this looks great, left minor comments about code duplication. Would also be nice to have a single package. 

(should address merge conflicts)

packages/trigger-client/src/TriggerClient.ts
yruej301: I think if we could share between engine and trigger service it's worth doing.

yruej301: Ah i see, i get it now. Tbh if input / output are different thinks make sense to leave it separate, my only concern was around the logic itself since its duplicated. But i prefer this over complicating the input / output types.

yruej301: Actually signing could deff be shared no?

yruej301: > Yeah it might be more readable to just duplicate

agree, plus we don't really have a great pattern for sharing a base across different clients and tbh i'd probably do the same thing we did here when it comes time to integrate this onto the python sdk

yruej301:
> @yruej301 can you give this one more look before i merge & publish?

yup, on it

yruej301 (approved):
new changes lgtm!

PR #113, opened 2023-07-09
Migrate to ethers v6

frankfka:
All sanity tests are ✅ 

Resource: https://docs.ethers.org/v6/migrating


apps/e2e/src/client/wsSanity.ts
```diff
@@ -19,7 +23,7 @@ async function wsSanity(context: RunContext) {
     expiration: nowInSeconds() + 60,
     // Limit price
     price: 28000,
-    amount: toFixedPoint(0.01),
+    amount: toFixedPoint(0.01).toString(),
```
yruej301, line 26: would it make sense to handle the conversion internally? or have a different version of `toFixedPoint` that returns a string?

packages/client/src/apis/market/MarketExecuteAPI.ts
```diff
@@ -23,7 +23,7 @@ export class MarketExecuteAPI extends BaseVertexAPI {
     const sender = await this.getChainSignerAddress();
 
     return this.context.engineClient.mintLp({
-      verifyingAddr: this.context.contracts.endpoint.address,
+      verifyingAddr: await this.getEndpointAddress(),
```
yruej301, line 26: what if we make the endpoint address a property on the client somewhere and only fetch it once on instantiation?

yruej301: ye, i think we should do that 👍

packages/contracts/src/common/types/subaccountTypes.ts
```diff
@@ -1,8 +1,6 @@
-import { Bytes } from 'ethers/lib/utils';
+export type SubaccountBytes32 = Uint8Array;
```
yruej301, line 1: @frankfka we can use `Bytes` from `'./bytes` here no?

packages/contracts/src/typechain-types/factories/Endpoint__factory.ts
```diff
@@ -2,8 +2,7 @@
 /* tslint:disable */
 /* eslint-disable */
 
-import { Contract, Signer, utils } from "ethers";
-import type { Provider } from "@ethersproject/providers";
+import { Contract, Interface, type ContractRunner } from "ethers";
```
yruej301, line 5: @frankfka curious why did the typechains got regenerated? to be compatible w/ v6?

packages/client/src/apis/subaccount/SubaccountExecuteAPI.ts
```diff
@@ -48,12 +48,18 @@ export class SubaccountExecuteAPI extends BaseVertexAPI {
     params: Pick<Subaccount, 'subaccountName'>,
   ) {
     const chainSigner = await this.getChainSigner();
-    const chainId = await chainSigner.getChainId();
+    const provider = chainSigner.provider;
```
yruej301, line 51: @frankfka can we use `getChainIdFromSigner` here?

packages/utils/src/math/bigDecimal.ts
```diff
@@ -1,5 +1,30 @@
...
+
 /**
- * Renamed `BigNumber` type from `bignumber.js`.
+ * BigDecimal is a renamed `BigNumber` type from `bignumber.js`. BigDecimalish is akin to `BigNumberish`, but
+ * includes valid values & instances for BigDecimal.
+ *
  * @see https://mikemcl.github.io/bignumber.js/
  */
-export { BigNumber as BigDecimal } from 'bignumber.js';
+export type BigDecimalish = BigDecimal | BigDecimal.Value | BigNumberish;
```
yruej301, line 13: @frankfka why this? 👀

yruej301: oh nvm, i see this was an existing thing.

yruej301 (commented):
@frankfka changes look great -- think we should address having the endpoint contract be part of the context, left some other minor comments but great stuff! v6 seems a lot cleaner. 

(let's also address merge conflicts, etc)

apps/e2e/src/client/wsSanity.ts
yruej301: ah ic, i think this is fine as is 👍

packages/client/src/apis/spot/SpotExecuteAPI.ts
```diff
@@ -47,6 +47,6 @@ export class SpotExecuteAPI extends BaseSpotAPI {
       config.token,
       this.context.signerOrProvider,
     );
-    return erc20.mint(erc20.signer.getAddress(), params.amount);
+    return erc20.mint(await this.getChainSignerAddress(), params.amount);
```
yruej301, line 50: @frankfka qq - the chain signer can be updated after initial creation of the client right? that's why `getChainSignerAddress` is async? (otherwise we could make it part of the context but if iirc you can update after creation -- just confirming)

yruej301 (approved):
@frankfka changes look great!

minor optim questions -- can we add `chainId` and/or chainSigner address be part of the context to get rid of async?

packages/client/src/apis/spot/SpotExecuteAPI.ts
yruej301: makes sense to me! i don't think neither of those get in the way of critical APIs anyways so we should be good.

PR #125, opened 2023-08-24
Add reduce only support

frankfka:


yruej301 (approved):
@frankfka lgtm (reduce-only currently turned off on testnet though)

PR #128, opened 2023-09-15
LBA

frankfka:


yruej301:
Sounds greaat, will do a first pass today!

packages/contracts/abis/IEndpoint.json
```diff
@@ -1,298 +1,97 @@
...
-        "type": "bytes[]"
-      }
-    ],
-    "name": "submitTransactionsChecked",
-    "outputs": [],
-    "stateMutability": "nonpayable",
-    "type": "function"
-  }
-]
+{
```
yruej301, line 1: this is wrong -- it seems to be replacing the correct IEndpoint.json (from vertex-core) with the stubbed out IEndpoint from the LBA -- my guess is we don't want the LBA endpoint interface.

packages/contracts/src/common/deployments/vrtx/deployment.arbitrumOne.json
```diff
@@ -0,0 +1,13 @@
+{
```
yruej301, line 1: we should probably remove this until it's actually deployed

packages/client/src/apis/vrtxToken/VrtxTokenExecuteAPI.ts
```diff
@@ -0,0 +1,127 @@
...
+    const { totalAmount, proof } =
+      await this.context.indexerClient.getTokenClaimProof({
+        epoch: 6,
+        address: await this.getChainSignerAddress(),
+      });
+
+    const airdropContract = this.context.contracts.vrtxAirdrop;
+
+    // If no amount is given, then the full amount available is claimed
+    const amountToClaim = await (async () => {
```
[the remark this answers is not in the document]
yruej301: @frankfka is it possible to pass `{ claimAll: false }`? i.e: no amount but set the flag to `false`? if not possible then we're good!

yruej301:
yup, reviewing now

yruej301 (approved):
changes look good!

PR #140, opened 2023-10-29
Support ARB rewards

frankfka:
Also unbreaks VRTX rewards testing


yruej301 (approved):
looks great!!

one minor thing, would probably be good to setup integration testing for this at some point

PR #143, opened 2023-10-31
chore: v1 engine endpoints

yruej301:


packages/engine-client/src/endpoints.ts
```diff
@@ -2,8 +2,8 @@ import { ChainEnv } from '@vertex-protocol/contracts';
 
 export const ENGINE_CLIENT_ENDPOINTS: Record<ChainEnv, string> = {
   local: 'http://localhost:80',
-  testnet: 'https://gateway.test.vertexprotocol.com',
-  mainnet: 'https://gateway.prod.vertexprotocol.com',
+  testnet: 'https://gateway.test.vertexprotocol.com/v1',
```
yruej301, line 5: @frankfka we've v1'd everything the SDK uses (still not on mainnet though)

PR #147, opened 2023-11-10
Migrate to subaccount snapshots query

frankfka:


yruej301 (approved):
looks good! took a bit to follow the `snapshotsBySubaccount` mapping but i think it's fine!

PR #151, opened 2023-11-27
Make `subaccountOwner`/`verifyingAddr`/`chainId` Optional for Market Execute Calls

frankfka:
This allows clients to pass in params that would otherwise require an `await` call. On the FE we can pre-load these for faster execution times.


packages/client/src/apis/market/types.ts
```diff
@@ -1,49 +1,55 @@
...
 import {
   QueryListTriggerOrdersParams,
   TriggerExecuteCancelOrdersParams,
   TriggerExecuteCancelProductOrdersParams,
   TriggerExecutePlaceOrderParams,
 } from '@vertex-protocol/trigger-client';
 
-type ClientOrderParams<T> = Omit<T, 'order'> & {
-  order: WithoutSubaccountOwner<EngineOrderParams>;
+// TODO This is currently used only for the market API, where speed is more important. This could eventually be used for other executes
```
frankfka, line 15: Leaving this for the future - imo a bit low priority for the other executes?

yruej301 (commented):
@frankfka overall looks good, left some minor comments / suggestions (prolly low pri). 

I would just say we should make sure all sanity checks are ran, etc. this PR highlights the need of more testing on the SDK, will prio it.

yruej301 (approved):
looks good!

PR #153, opened 2023-12-05
chore: sunset arbitrum goerli

yruej301:


packages/contracts/update-bindings.sh
frankfka, line 1: I think this file is out of date right? maybe we just nuke it & the npm script?

frankfka (approved):

packages/contracts/update-bindings.sh
yruej301: yup, I will nuke it

yruej301: looks like the npm script is already removed but nuked the file

PR #157, opened 2023-12-30
Add ability to claim & stake VRTX / Staking rewards

frankfka:


packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -81,12 +82,11 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
       return availableAmount.toFixed();
     })();
 
-    return this.context.contracts.vrtxAirdrop.claim(
-      params.epoch,
-      amountToClaim,
-      totalAmount.toFixed(),
-      proof,
-    );
+    const claimFn = params.stake
```
frankfka, line 85: Thoughts on this, versus exposing a new fn? there was a decent bit of logic up above that i wanted to share

frankfka: Fair enough

yruej301 (approved):
lgtm!

PR #159, opened 2024-01-05
Bound margin usage fractions between 0 and 1

frankfka:
I found a case where initial/maint margin usage could be negative (which makes no sense), if the inital/maint/unweighted healths are negative - i.e. account is already underwater.

In the same change, im bounding margin usage from 0->100% (this is currently done on the FE already, but we reduce duplication by putting it in the SDK)

### DIFF
--- a/packages/contracts/src/utils/subaccountInfo/subaccountInfo.ts
+++ b/packages/contracts/src/utils/subaccountInfo/subaccountInfo.ts
@@ -116,7 +116,8 @@ export interface MarginUsageFractions {
 }
 
 /**
- * Calculate margin usage fractions. Which is = (unweighted health - initial/maint health) / unweighted health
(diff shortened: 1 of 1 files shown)


packages/contracts/src/utils/subaccountInfo/subaccountInfo.ts
```diff
@@ -162,33 +163,14 @@ export function calcSubaccountMarginUsageFractions(
...
-  };
-}
-
-/**
- * Calculates margin usage fractions for a single balance.
- *
- * @param balance
- * @param summary
- */
-export function calcBalanceMarginUsageFractions(
```
frankfka, line 177: This isn't used on the FE and i think it's using an old calculation

yruej301 (approved):
LGTM!

PR #164, opened 2024-01-31
chore: add trigger order id

yruej301:


packages/trigger-client/src/TriggerClient.ts
```diff
@@ -83,6 +83,7 @@ export class TriggerClient {
     );
 
     const executeParams: TriggerServerExecuteRequestByType['place_order'] = {
+      id: params.id?.toNumber() || null,
```
frankfka, line 86: We should be using `??` (can we adjust on engine if its not being used there?)

yruej301: oh ic, think we are not using there will fix. what's the diff again? 😭

yruej301: oh nvm we are!

frankfka: https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or

yruej301: oh TIL!

yruej301: (actually you might have point this out in the past but think i won't forget again!)

packages/trigger-client/src/TriggerClient.ts
```diff
@@ -83,6 +83,7 @@ export class TriggerClient {
     );
 
     const executeParams: TriggerServerExecuteRequestByType['place_order'] = {
+      id: params.id?.toNumber() ?? null,
```
frankfka, line 86: We need to be clear that consumers can't specify anything greater than max safe int - otherwise `toNumber()` will fail and just give an incorrect ID. Ideally BE can parse both strings and numbers so we can just convert to string here

frankfka (approved):

frankfka (approved):

PR #165, opened 2024-02-07
Vertex V1 / V2 Interop Support

frankfka:
@yruej301 can you advise on the new fields for V2?


packages/indexer-client/src/types/VertexTx.ts
```diff
@@ -11,7 +11,10 @@ export interface VertexLiquidateSubaccountTx {
     sender: string;
     liquidatee: string;
     mode: number;
-    health_group: number;
+    // On V2 - should encode health group
+    product_id: number | undefined;
```
frankfka, line 15: Also - confirming that these are still u32 and coming back as numbers and not strings?

yruej301: will confirm!

yruej301: it's a number

frankfka: Okay this should be good for now then 👍🏻 but any other fields that should be added / removed?

PR #166, opened 2024-02-08
blast sepolia support

yruej301:


packages/client/src/context.ts
```diff
@@ -133,6 +133,26 @@ export async function createClientContext(
           triggerEndpoint: TRIGGER_CLIENT_ENDPOINTS.mainnet,
         };
       }
+      if (opts === 'blastTestnet') {
```
frankfka, line 136: Do you have bandwidth to clean this up a bit? This fn is quite large and there's a decent bit of duplication

frankfka (approved):

packages/client/src/context.ts
yruej301: not immediately, but can add a backlog item to address.

PR #167, opened 2024-02-10
Optional Chain ID for executes on Vertex Client

frankfka:
I left a TODO originally to do this so finally got around to it. Had to rename the interfaces for EIP712 as they were conflicting with the new types in `packages/client`

### DIFF
--- a/packages/client/src/apis/market/MarketExecuteAPI.ts
+++ b/packages/client/src/apis/market/MarketExecuteAPI.ts
@@ -1,31 +1,27 @@
 import { BaseVertexAPI } from '../base';
 import {
+  BurnLpParams,
   CancelAndPlaceOrderParams,
   CancelOrdersParams,
   CancelProductOrdersParams,
   CancelTriggerOrdersParams,
   CancelTriggerProductOrdersParams,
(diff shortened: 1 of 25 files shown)


packages/client/src/apis/market/MarketExecuteAPI.ts
```diff
@@ -1,31 +1,27 @@
...
    * @param params
    */
-  async mintLp(params: OptionalSubaccountOwner<EngineExecuteMintLpParams>) {
+  async mintLp(params: MintLpParams) {
     return this.context.engineClient.mintLp({
       ...params,
-      chainId: await this.getSignerChainId(),
-      verifyingAddr: this.getEndpointAddress(),
+      chainId: await this.getSignerChainIdIfNeeded(params),
+      verifyingAddr: params.verifyingAddr ?? this.getEndpointAddress(),
```
frankfka, line 24: There isn't much benefit to passing this in, but I added so that we can leverage the same `OptionalSignatureParams` type

yruej301 (approved):
these makes the execute types so much cleaner, makes a lot of sense to me 👍

PR #168, opened 2024-02-12
Fix RangeError on some wallets when trying to claim rewards

frankfka:
I have no idea why this happens, but this _should_ fix: https://vertex-xwn1857.slack.com/archives/C054VJ2DEFL/p1707739663393299

(For context, user resolved by using metamask)


yruej301 (approved):
send it

PR #169, opened 2024-02-18
Add pagination to rewards query

frankfka:


packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -623,8 +625,8 @@ export class IndexerBaseClient {
    * @param params
    */
   async getSubaccountArbRewards(
```
yruej301, line 627: should this be `getArbRewards`

yruej301 (commented):
the core pagination logic looks good, just left a minor comment

packages/indexer-client/src/IndexerBaseClient.ts
frankfka: Yeah good call!

PR #171, opened 2024-02-19
Cleanup Vertex Client Context Creation

frankfka:


packages/client/src/context.ts
```diff
@@ -50,22 +49,7 @@ export interface VertexClientContext {
  * Args for creating a context
  */
 interface VertexClientContextOpts {
-  // Contract addresses
-  contracts: {
-    // Only the querier is required, but you can specify all the contracts to minimize RPC calls
```
frankfka, line 55: I don't think there's much benefit in allowing this - most people are just going to use testnet/mainnet/etc.

yruej301 (approved):
the simplification looks good

PR #175, opened 2024-02-28
Standardize interface naming for indexer types

frankfka:


packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -103,8 +103,8 @@ export class IndexerBaseClient {
    * @param params
    */
   async listSubaccounts(
```
yruej301, line 105: minor but how about we just call this `getSubaccounts` and rename params/response to just `IndexerQuerySubaccountsParams` / `IndexerQuerySubaccountsResponse`. seems more standard.

packages/indexer-client/src/IndexerClient.ts
```diff
@@ -7,36 +7,36 @@ import { toBigDecimal } from '@vertex-protocol/utils';
...
+  IndexerQuerySubaccountSettlementEventsResponse,
   IndexerSettlementEvent,
   PaginatedIndexerEventsResponse,
 } from './types';
 
 export class IndexerClient extends IndexerBaseClient {
   async getPaginatedSubaccountMatchEvents(
-    params: GetIndexerSubaccountMatchEventParams,
-  ): Promise<GetIndexerSubaccountMatchEventsResponse> {
+    params: IndexerQuerySubaccountMatchEventParams,
```
yruej301, line 38: it seems some params / response include `paginated` e.g: `IndexerQueryPaginatedOrdersParams` and some don't, think we should standardize it.

packages/indexer-client/src/IndexerClient.ts
```diff
@@ -183,8 +183,8 @@ export class IndexerClient extends IndexerBaseClient {
   }
 
   async getPaginatedSubaccountOrders(
-    params: GetIndexerPaginatedOrdersParams,
-  ): Promise<GetIndexerPaginatedOrdersResponse> {
+    params: IndexerQueryPaginatedOrdersParams,
```
yruej301, line 186:
```suggestion
    params: IndexerQuerySubaccountOrdersParams,
```

packages/indexer-client/src/IndexerClient.ts
```diff
@@ -183,8 +183,8 @@ export class IndexerClient extends IndexerBaseClient {
...
+  ): Promise<IndexerQueryPaginatedOrdersResponse> {
```
yruej301, line 187:
```suggestion
  ): Promise<IndexerQuerySubaccountOrdersResponse> {
```

yruej301 (commented):
overall the renaming looks good, left some minor comments

packages/indexer-client/src/IndexerClient.ts
frankfka: Hm these are paginated queries with pagination `meta`, but the queries on `IndexerBaseClient` don't really have pagination built in

PR #177, opened 2024-02-28
Standardize interface naming for engine and client types

frankfka:


packages/client/src/apis/market/MarketQueryAPI.ts
```diff
@@ -76,7 +76,7 @@ export class MarketQueryAPI extends BaseVertexAPI {
    * @description Queries the offchain trigger service to list trigger orders. Requires a signature
    * @param params
    */
-  async getTriggerOrders(params: GetTriggerOrdersParams) {
+  async getTriggerOrders(params: QueryListTriggerOrdersParams) {
```
yruej301, line 79:
```suggestion
  async getTriggerOrders(params: QueryTriggerOrdersParams) {
```

packages/client/src/apis/market/MarketQueryAPI.ts
```diff
@@ -50,15 +50,15 @@ export class MarketQueryAPI extends BaseVertexAPI {
   /**
    * @description Queries engine to determine if the order can be submitted within health requirements
    */
-  async validateOrderParams(params: ValidateEngineOrderParams) {
+  async validateOrderParams(params: EngineQueryValidateOrderParams) {
```
yruej301, line 53: would it be possible to drop the `Engine` prefix in all these? seems it would be more standard with the rest

packages/client/src/apis/market/types.ts
```diff
@@ -6,51 +6,52 @@ import {
...
 >;
 
-export type BurnLpParams = OptionalSignatureParams<
+export type ExecuteBurnLpParams = OptionalSignatureParams<
   OptionalSubaccountOwner<EngineExecuteBurnLpParams>
 >;
 
-export type GetTriggerOrdersParams =
-  OptionalSignatureParams<QueryListTriggerOrdersParams>;
+export type QueryListTriggerOrdersParams =
```
yruej301, line 56: re: `TriggerQueryListTriggerOrdersParams` -- think we should find a better name for this. why can't this just be `QueryTriggerOrdersParams`. in the backend we have the concept of `Signed` vs `Unsigned` stuff, maybe we could do something similar here if the diff is just an optional signature.

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -73,7 +73,7 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
    *
    * @param params
    */
-  async claimAndStakeLiquidTokens(params: ClaimLiquidTokensParams) {
+  async claimAndStakeLiquidTokens(params: ExecuteClaimLiquidTokensParams) {
```
yruej301, line 76: should this be `ClaimAndStackLiquidTokensParams`? why do we need `Execute` prefix here?

yruej301 (commented):
ngl I am not convinced this is better tbh. It would be much nicer if things weren't as wordy as they are, like imo --

- `GetSomethingParams` >> `QuerySomethingParams`
- `ThisActionParams` >> `ExecuteThisActionParams`

e.g: `GetNoncesParams` instead of `QueryNoncesParams` and `PlaceOrderParams` instead of `ExecutePlaceOrderParams`

i think should we just try to simplify these names.

I know we kinda agreed to this, but seeing it now think we can do better by simplifying these names. literally for every `Query` let's just do `GetTheThing` and for every execute we don't need to include `Execute` in the name. 

and for responses just like `NoncesResponse`, `IpBlockResponse`, etc

packages/client/src/apis/market/types.ts
frankfka: `TriggerQueryListTriggerOrdersParams` - the `Trigger` prefix is to indicate that this is talking to the the trigger service. It's clear enough with this specific case but for example - `GetIndexerOrdersParams` without `Indexer` would be `GetOrdersParams`: would this be talking to the engine or the indexer?

packages/client/src/apis/market/MarketQueryAPI.ts
frankfka: Then we need to redeclare something like `type ValidateOrderParams = EngineQueryValidateOrderParams`

frankfka:
> ngl I am not convinced this is better tbh. It would be much nicer if things weren't as wordy as they are, like imo --
> 
>     * `GetSomethingParams` >> `QuerySomethingParams`
> 
>     * `ThisActionParams` >> `ExecuteThisActionParams`
> 
> 
> e.g: `GetNoncesParams` instead of `QueryNoncesParams` and `PlaceOrderParams` instead of `ExecutePlaceOrderParams`
> 
> i think should we just try to simplify these names.
> 
> I know we kinda agreed to this, but seeing it now think we can do better by simplifying these names. literally for every `Query` let's just do `GetTheThing` and for every execute we don't need to include `Execute` in the name.
> 
> and for responses just like `NoncesResponse`, `IpBlockResponse`, etc

Would `Get` be preferable to `Query`? I think I landed on the `Query` prefix because our endpoints are either `/query` or `/execute` but honestly just having things standardized would be a value gain.

I feel like we do need `Get` / `Query` prefixes on responses though - otherwise `GetEngineLinkedSignerResponse` -> `EngineLinkedSignerResponse`. Just by reading the type name, is it from an execute or a query?

yruej301:
@frankfka  how about this -- basically we keep client naming with the least verbosity possible while server naming can contain the service name and we keep `Query/Execute` as well. lmk

# Engine
## Queries
### Client
- `GetItemParams` e.g `GetNoncesParams`
- `ItemResponse` e.g `NoncesResponse`

### Server
- `QueryEngineItemParams` eg.: `QueryEngineNoncesParams`
- `EngineItemResponse` eg.: `EngineNoncesResponse`

## Executes
### Client
- `ActionParams` e.g: `PlaceOrderParams`
- `ActionResponse` e.g: `PlaceOrderResponse`
### Server
- `ExecuteEngineActionParams` e.g: `ExecuteEnginePlaceOrderParams`
- `ExecuteEngineActionResponse` e.g: `ExecuteEnginePlaceOrderResponse` (if needed)

# Indexer
## Queries
### Client
- `GetItemParams` or `GetHistoricalItemParams` (when name clashes) e.g: `GetSubaccountsParams` or `GetHistoricalOrdersParams`
- `ItemParams` or `HistoricalItemParams` (when name clashes) e.g: `SubaccountsResponse` or `HistoricalOrdersResponse`
### Server
- `QueryIndexerItemParam` e.g: `QueryIndexerSubaccountsParams`
- `IndexerItemResponse` e.g: `IndexerSubaccountsResponse`

# Trigger
## Queries
### Client
- `GetTriggerItemParams` e.g: `GetTriggerOrderParams`
- `TriggerItemResponse` e.g: `TriggerOrdersResponse`
### Server
- `QueryTriggerItemParams` e.g: `QueryTriggerOrdersParams`
- `TriggerItemResponse` (think we can tweak this if clashing between client / server)

## Executes
### Client
- `TriggerActionParams` e.g: `PlaceTriggerOrderParams` or `CancelTriggerOrderParams`
- `TriggerActionResponse` e.g: `TriggerOrderResponse` or `PlaceTriggerOrderResponse`

### Server
- `ExecuteTriggerActionParams` e.g: `ExecutePlaceTriggerOrderParams`
- `ExecuteTriggerActionResponse` e.g: `ExecutePlaceTriggerOrderResponse`

frankfka:
@yruej301 so is the general philosophy that we have `GetXXX` for client types but `QueryXXX` for backend types? Would this be confusing for users as the distinction between `Get` and `Query` is not that clear?

Additionally, how would you handle types for the "top level" client? See `PlaceOrderParams`

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -102,19 +101,19 @@ export class EngineQueryClient extends EngineBaseClient {
...
-    params: GetEngineEstimatedSubaccountSummaryParams,
-  ): Promise<GetEngineSubaccountSummaryResponse> {
+    params: EngineQueryEstimatedSubaccountSummaryParams,
+  ): Promise<EngineQuerySubaccountSummaryResponse> {
     const subaccount = subaccountToHex({
       subaccountOwner: params.subaccountOwner,
       subaccountName: params.subaccountName,
     });
-    const queryParams: EngineServerSubaccountInfoQueryParams = {
+    const queryParams: EngineServerQuerySubaccountInfoParams = {
```
yruej301, line 110:
```suggestion
    const queryParams: ServerQueryEngineSubaccountInfoParams = {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -102,19 +101,19 @@ export class EngineQueryClient extends EngineBaseClient {
...
       subaccount: subaccount,
       txns: params.txs.map(
         (
           tx,
         ): NonNullable<
-          EngineServerSubaccountInfoQueryParams['txns']
+          EngineServerQuerySubaccountInfoParams['txns']
```
yruej301, line 116:
```suggestion
          ServerQueryEngineSubaccountInfoParams['txns']
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -162,8 +161,8 @@ export class EngineQueryClient extends EngineBaseClient {
    * @param params
    */
   async getSymbols(
-    params: GetEngineSymbolsParams,
-  ): Promise<EngineSymbolsResponse> {
+    params: EngineQuerySymbolsParams,
```
yruej301, line 164:
```suggestion
    params: QueryEngineSymbolsParams,
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -162,8 +161,8 @@ export class EngineQueryClient extends EngineBaseClient {
...
+  ): Promise<EngineQuerySymbolsResponse> {
```
yruej301, line 165:
```suggestion
  ): Promise<EngineSymbolsResponse> {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -176,7 +175,7 @@ export class EngineQueryClient extends EngineBaseClient {
   /**
    * Retrieves all market states as per the offchain engine. Same return interface as contracts
    */
-  async getAllMarkets(): Promise<GetEngineAllMarketsResponse> {
+  async getAllMarkets(): Promise<EngineQueryAllMarketsResponse> {
```
yruej301, line 178:
```suggestion
  async getAllMarkets(): Promise<QueryEngineAllMarketsResponse> {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -193,7 +192,7 @@ export class EngineQueryClient extends EngineBaseClient {
   /**
    * Retrieves all health groups (linked spot & perp products) from the engine
    */
-  async getHealthGroups(): Promise<GetEngineHealthGroupsResponse> {
+  async getHealthGroups(): Promise<EngineQueryHealthGroupsResponse> {
```
yruej301, line 195:
```suggestion
  async getHealthGroups(): Promise<EngineHealthGroupsResponse> {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -214,8 +213,8 @@ export class EngineQueryClient extends EngineBaseClient {
    * @param params
    */
   async getOrder(
-    params: GetEngineOrderParams,
-  ): Promise<GetEngineOrderResponse> {
+    params: EngineQueryOrderParams,
```
yruej301, line 216:
```suggestion
    params: QueryEngineOrderParams,
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -214,8 +213,8 @@ export class EngineQueryClient extends EngineBaseClient {
...
+  ): Promise<EngineQueryOrderResponse> {
```
yruej301, line 217:
```suggestion
  ): Promise<EngineOrderResponse> {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -230,8 +229,8 @@ export class EngineQueryClient extends EngineBaseClient {
    * @param params
    */
   async validateOrderParams(
-    params: ValidateEngineOrderParams,
-  ): Promise<ValidateEngineOrderResponse> {
+    params: EngineQueryValidateOrderParams,
```
yruej301, line 232:
```suggestion
    params: QueryEngineValidateOrderParams,
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -230,8 +229,8 @@ export class EngineQueryClient extends EngineBaseClient {
...
+  ): Promise<EngineQueryValidateOrderResponse> {
```
yruej301, line 233:
```suggestion
  ): Promise<EngineValidateOrderResponse> {
```

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -253,8 +252,8 @@ export class EngineQueryClient extends EngineBaseClient {
    * @param params
    */
   async validateSignedOrderParams(
-    params: ValidateSignedEngineOrderParams,
-  ): Promise<ValidateEngineOrderResponse> {
+    params: EngineQueryValidateSignedOrderParams,
```
yruej301, line 255: same for the rest of these

packages/engine-client/src/EngineWebClient.ts
```diff
@@ -23,7 +26,7 @@ export class EngineWebClient extends EngineBaseClient {
           if (res.status !== 403) {
             return false;
           }
-          const resData: EngineServerIpBlockResponse = res.data;
+          const resData: EngineServerQueryIpBlockResponse = res.data;
```
yruej301, line 29:
```suggestion
          const resData: EngineIpBlockResponse = res.data;
```

packages/engine-client/src/EngineWebClient.ts
```diff
@@ -56,7 +59,7 @@ export class EngineWebClient extends EngineBaseClient {
   /**
    * Retrieves current server epoch in milliseconds
    */
-  async getTime(): Promise<GetEngineTimeResponse> {
+  async getTime(): Promise<EngineQueryTimeResponse> {
```
yruej301, line 62:
```suggestion
  async getTime(): Promise<EngineTimeResponse> {
```

packages/engine-client/src/types/serverQueryTypes.ts
```diff
@@ -144,28 +143,28 @@ export type EngineServerQueryRequest<
   type: TRequestType;
 } & EngineServerQueryRequestByType[TRequestType];
 
-export interface EngineServerContractsResponse {
+export interface EngineServerQueryContractsResponse {
```
yruej301, line 146:
```suggestion
export interface ServerEngineContractsResponse {
```

frankfka:
@yruej301

frankfka:
@yruej301 

`for executes: drop Execute prefix` - this is fine 

`for query params: switch from EngineQueryThisParams to QueryEngineThisParams` - What exactly is the value gain of this?

`for query responses: drop Query i.e: EngineQueryThisResponse to EngineThisResponse` - Doesn't this break the consistency with naming the params with response? what's the benefit of dropping `Query`? (I suppose less verbose?)

**Server Types**
- Are we dropping `Execute` as well? Are we dropping `Query`?
- Does `EngineServer` vs `ServerEngine` make a difference?

frankfka:
https://github.com/vertex-protocol/vertex-typescript-sdk/pull/181

yruej301:
@frankfka just for the sake of clarifying where i was going with here --

> for query params: switch from EngineQueryThisParams to QueryEngineThisParams - What exactly is the value gain of this?

 the benefit i saw is that we can standardize all queries across the board `Query[Engine|Indexer|Trigger|Web|""]...` and imo `QueryEngine...` is more clear than `EngineQuery`

> for query responses: drop Query i.e: EngineQueryThisResponse to EngineThisResponse - Doesn't this break the consistency with naming the params with response? what's the benefit of dropping Query? (I suppose less verbose?)

I don't think query params / response need to follow the same pattern with each other, but i do think all queries across the board and all responses across the board should follow the same pattern. 

I think if we did --
- Query params: `QueryEngineThisResponse`
- Query response: `EngineThisResponse`

^seems both consistent and clear to me

yruej301:
> #181

yup, this looks good!

PR #181, opened 2024-03-17
Standardize SDK interface naming

frankfka:


packages/engine-client/src/types/clientExecuteTypes.ts
```diff
@@ -87,7 +86,7 @@ export interface EngineExecuteRequestParamsByType {
   withdraw_collateral: EngineExecuteWithdrawCollateralParams;
   burn_lp: EngineExecuteBurnLpParams;
   place_order: EngineExecutePlaceOrderParams;
-  cancel_orders: EngineExecuteCancelOrdersParams;
+  cancel_orders: EngineCancelOrdersParams;
```
frankfka, line 89: For client types - dropping `Execute`

packages/engine-client/src/types/serverQueryTypes.ts
```diff
@@ -310,11 +309,11 @@ export interface EngineServerQueryResponseByType {
   symbols: EngineServerSymbolsResponse;
   all_products: EngineServerAllProductsResponse;
   health_groups: EngineServerHealthGroupsResponse;
-  order: EngineServerGetOrderResponse;
+  order: EngineServerOrderResponse;
   orders: EngineServerProductOrdersResponse;
   validate_order: EngineServerValidateOrderResponse;
   subaccount_orders: EngineServerSubaccountOrdersResponse;
-  fee_rates: EngineServerSubaccountFeeRatesResponse;
+  fee_rates: EngineServerFeeRatesResponse;
```
frankfka, line 316: Keeping the `EngineServer` prefix, but naming is consistent with the name of the backend query (ex. `FeeRates`, `Order`, `HealthGroups`)

frankfka (commented):
Keeping the `GetEngineOrderParams/Response` - I think this is okay as the function is named `getOrder`, so the interface naming is close to the function name, but we still "namespace" by `Engine`

packages/engine-client/src/types/serverSubscriptionEventTypes.ts
```diff
@@ -3,7 +3,7 @@ import { EngineServerPriceTickLiquidity } from './serverQueryTypes';
 /**
  * @description Event from subscribing to a `trade` stream.
  */
-export interface SubscriptionTradeEvent {
+export interface EngineServerSubscriptionTradeEvent {
```
frankfka, line 6: I'd like to prefix these with `EngineServer` to standardize with query / executes

packages/engine-client/src/types/clientExecuteTypes.ts
yruej301: sounds good!

packages/engine-client/src/types/serverSubscriptionEventTypes.ts
yruej301: yea, i think change looks good as well 👍

yruej301:
@frankfka i think the proposed changes here are good

packages/trigger-client/src/types/clientTypes.ts
```diff
@@ -109,14 +101,6 @@ export interface TriggerOrderInfo {
   updatedAt: number;
 }
 
-export interface QueryListTriggerOrdersResponse {
+export interface TriggerListOrdersResponse {
   orders: TriggerOrderInfo[];
 }
-
-export interface TriggerQueryRequestParamsByType {
```
frankfka, line 116: These just added complexity imo - not really needed

frankfka:
I think everything should be addressed now. Would be useful to pull this branch and go through locally.

I skipped renaming from `EngineQuery` -> `QueryEngine` - this would be a pretty big change and i don't feel like the value is there.

Also skipped adding `Query` to indexer server types. Given that indexer only has queries I think we're safe to skip

PR #185, opened 2024-04-05
Blitz Points

frankfka:


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -542,6 +542,41 @@ export type GetIndexerClaimArbMerkleProofsParams =
...
 
+/**
+ * Blitz points
+ */
+
+export interface GetIndexerBlitzPointsParams {
+  address: string;
+}
+
+export interface GetIndexerBlitzPointsResponse {
```
yruej301, line 553: is it possible to add comments for what each of these fields are?

yruej301 (commented):
just realized this is a minor PR. left a comment but would it be possible to also add some sanity checks?

packages/indexer-client/src/types/clientTypes.ts
frankfka: Done!

apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -84,26 +84,43 @@ async function fullSanity(context: RunContext) {
...
     limit: 1,
-    productIds: [1, 2, 3, 4],
+    productIds: [2, 3, 4],
   });
 
   prettyPrint('Market snapshots', marketSnapshots);
 
   const multiProductSnapshots = await client.getMultiProductSnapshots({
-    productIds: [1, 2, 3],
+    productIds: [2, 3],
```
frankfka, line 123: Picking product IDs that work on both blitz & vertex

yruej301 (approved):
lgtm!

PR #187, opened 2024-04-27
Tighten generic typing for engine execute result

frankfka:
Fixes [this](https://github.com/vertex-protocol/vertex-web-monorepo/pull/2053#discussion_r1581786197), now result types are "tightened" to the specific request type, now we can properly access the `data` field without casting:

image


yruej301 (approved):
lgtm!

PR #188, opened 2024-04-29
Cleanup: Remove getOrderDigest from EngineClient

frankfka:
We pretty much just forwarded arguments to the same util function, so was thinking that we don't need this on the client itself so that we have one "source of truth" function for computing digest

### DIFF
--- a/apps/e2e/src/client/fullSanity.ts
+++ b/apps/e2e/src/client/fullSanity.ts
@@ -1,13 +1,14 @@
-import { RunContext } from '../utils/types';
+import { createVertexClient, PlaceOrderParams } from '@vertex-protocol/client';
 import {
   getChainIdFromSigner,
+  getOrderDigest,
   getOrderNonce,
 } from '@vertex-protocol/contracts';
-import { toBigDecimal, toFixedPoint } from '@vertex-protocol/utils';
-import { runWithContext } from '../utils/runWithContext';
-import { createVertexClient, PlaceOrderParams } from '@vertex-protocol/client';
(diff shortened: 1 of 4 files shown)


yruej301 (commented):
wdyt about doing this the other way around and leave `vertexClient.context.engineClient.getOrderDigest` as an engine client specific util (similar to `vertexClient.context.engineClient.getOrderbookAddress`) thinking may be best for discoverability for API users

frankfka:
@yruej301 I think `getOrderbookAddress` makes sense given it needs the query client to make the `contracts` query. We could just leave it as-is though if you think it helps with discoverability. My original thought is that a util fn not requiring setup of the client is generally better left as just a util function, as there's less overhead to using the function itself (importing & using it versus needing to setup a `VertexClient` to do so

yruej301:
> @yruej301 I think `getOrderbookAddress` makes sense given it needs the query client to make the `contracts` query. We could just leave it as-is though if you think it helps with discoverability. My original thought is that a util fn not requiring setup of the client is generally better left as just a util function, as there's less overhead to using the function itself (importing & using it versus needing to setup a `VertexClient` to do so

i think this makes sense given we do something similar with `getOrderNonce`, ideally we just have this in one place and i get why we need them in `contracts`. maybe we can do another pass to move them as util but inside the `engine-client` package

yruej301 (approved):
@frankfka i think this is fine. think we should eventually move `getOrderDigest` and `getOrderNonce` as utils inside the `engine-client` package.

PR #189, opened 2024-04-29
Rename chainenv to be arbitrum-agnostic

frankfka:
This will also help with chain switching on the FE, ideally we have some sort of query parameter that lets us point users to a specific chain on-load. For example: `app.vertexprotocol.com/portfolio/overview?chain=arbitrum` 

Also removed the `Mainnet` postfix. This is to align more with other APIs (ex. [Viem](https://github.com/wagmi-dev/viem/tree/main/src/chains/index.ts) doesn't have mainnet postfixes) and to make the URLs a bit nicer


yruej301 (approved):
lgtm!

PR #190, opened 2024-05-01
Remove failure result from execute promises as errors are thrown and not returned

frankfka:
Title is self-explanatory, we never actually return the failure result as it gets thrown


yruej301 (approved):
@frankfka changes seem fine, would be good to add to sanity tests

PR #192, opened 2024-05-03
add multi ts product snaps support

clarkohw:


apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -108,6 +108,18 @@ async function fullSanity(context: RunContext) {
 
   prettyPrint('Multiple products snapshots', multiProductSnapshots);
 
+  const now = Math.round(Date.now() / 1000);
+  const multiTimestampProductSnapshots =
+    await client.getMultiTimestampProductSnapshots({
+      productIds: [1, 2, 3],
+      timestamps: [now, now - 3600, now - 86400],
```
frankfka, line 115: fyi we have `TimeInSeconds`

apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -108,6 +108,18 @@ async function fullSanity(context: RunContext) {
...
+    });
+
+  prettyPrint(
+    'Multiple timestamp and product products snapshots',
```
frankfka, line 119: typo?

packages/indexer-client/src/types/serverTypes.ts
```diff
@@ -268,11 +268,10 @@ export interface IndexerServerProductsResponse {
   txs: IndexerServerTx[];
 }
 
-// Map of productId -> IndexerServerProductSnapshot
-export type IndexerServerMultiProductsResponse = Record<
-  string,
-  IndexerServerProductSnapshot
->;
+// Map of productId -> IndexerServerProductSnapshot or timestamp -> (productID -> IndexerServerProductSnapshot)
+export type IndexerServerMultiProductsResponse =
```
frankfka, line 272: Let's simplify this and just force users to use `max_time?: number[]` - it makes the processing a lot easier without loss of functionality given that using `number[]` is a superset of passing a single `number`

apps/e2e/src/indexer-client/fullSanity.ts
clarkohw: actually not, but i does look weird, can update

packages/indexer-client/src/types/serverTypes.ts
clarkohw: while `number[]` is a super set, the response is different if an array is passed vs a single `number`. if its an array of 1 element, it will still give the map of `timestamp` -> `productId -> snapshots`. we could process the single number array into `number`  in `IndexerClient` though if we want to make this change. the intention is to make sure this doesn't break the current multiTimestamp query (in the sdk)

frankfka: It won't break unless users update, but if we have the invariant that the param is always `number[]` then the response will always be the new variant and not the old

clarkohw: got it. is it not an issue for FE, since it uses the existing query, and will want to update to the new one? or i guess it will be an intentional upgrade so any issues will be fixed right?

frankfka: Yeah we have to bump the package version for this to take effect, so it all goes into one PR

packages/indexer-client/src/types/serverTypes.ts
```diff
@@ -268,10 +268,10 @@ export interface IndexerServerProductsResponse {
   txs: IndexerServerTx[];
 }
 
-// Map of productId -> IndexerServerProductSnapshot
+// Map of productId -> IndexerServerProductSnapshot or timestamp -> (productID -> IndexerServerProductSnapshot)
```
frankfka, line 271: Now its' just `timestamp -> (productID -> IndexerServerProductSnapshot)` right?

frankfka (approved):
Let's hold off on merging - will make a FE task for this. Thanks!

PR #193, opened 2024-05-06
Upgrade Ethers to 6.12.1

frankfka:


package.json
```diff
@@ -36,5 +36,8 @@
     "typedoc-plugin-extras": "^2.3.2",
     "typedoc-plugin-missing-exports": "^1.0.0",
     "typescript": "^5.0.3"
+  },
+  "resolutions": {
```
frankfka, line 40: This allows us to specify a single version across all usecases

PR #194, opened 2024-05-07
chore: add slow-mode withdrawal sample

yruej301:


yruej301:
@frankfka had a ton of back and forth with a discord user on how to do this, so i just added this example so we can share it easily with them. confirmed this works. took it from your PR https://github.com/vertex-protocol/vertex-web-monorepo/pull/884/files!

frankfka:
Should this be part of sanity (hesitant to make this larger) or part of some `examples/` directory where we could maintain scripts that do smaller chunks of tasks? For example:
- Deposit & withdraw
- Place trade
- Place trigger order
- etc

yruej301:
> Should this be part of sanity (hesitant to make this larger) or part of some `examples/` directory where we could maintain scripts that do smaller chunks of tasks? For example:
> 
> * Deposit & withdraw
> * Place trade
> * Place trigger order
> * etc

@frankfka initially had the same thought but we've been using our sanity checks as examples at this point and we already cover pretty much everything else except for this specific example. don't think worth duplicating all the other stuff like place trade, etc.

frankfka:
> > Should this be part of sanity (hesitant to make this larger) or part of some `examples/` directory where we could maintain scripts that do smaller chunks of tasks? For example:
> > 
> > * Deposit & withdraw
> > * Place trade
> > * Place trigger order
> > * etc
> 
> @frankfka initially had the same thought but we've been using our sanity checks as examples at this point and we already cover pretty much everything else except for this specific example. don't think worth duplicating all the other stuff like place trade, etc.

I think what we could do for multi-line examples is split out a `utils` folder of sorts, then we can call the util fn within sanity tests if we want to include those in the e2e testing

frankfka (approved):
For now though we're good to merge this, maybe a TODO in the code is sufficient for us to follow up at some point re: utils

PR #196, opened 2024-05-15
Remove data response from trigger execute result

frankfka:
Context: https://vertex-xwn1857.slack.com/archives/C05914S2D3P/p1715801548802869

Ideally we can add this, but for now, fixing SDK


yruej301 (approved):
gotcha, fix looks good

PR #199, opened 2024-05-30
Add React Client Package

frankfka:
Extracted `@vertex-protocol/web-data` from the web monorepo into this SDK repo. Ideally, I would have liked to publish the package directly from the web monorepo, but was having issues with build. This seems to work fine though!


lerna.json
```diff
@@ -3,5 +3,5 @@
   "useNx": true,
   "useWorkspaces": true,
   "npmClient": "yarn",
-  "version": "0.34.0"
+  "version": "0.34.2"
```
frankfka, line 6: Had to publish for testing

packages/web-data/package.json
frankfka, line 1: This file is important to take a look at

packages/web-data/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "build": "tsc -b",
+    "dev": "tsc -w"
+  },
+  "dependencies": {
+    "d3-format": "^3.1.0"
+  },
+  "peerDependencies": {
+    "@tanstack/react-query": "^5",
```
frankfka, line 35: More permissive versioning here for consumers. `devDependencies` is what's required for local dev on the SDK

packages/web-data/src/context/evm/CoreEVMContextProvider.tsx
frankfka, line 1: These other files are copied over, so not sure if you need to check

packages/web-data/README.md
```diff
@@ -0,0 +1 @@
+# `@vertex-protocol/web-data`
```
yruej301, line 1: a really nice to have but would be nice to show how a frontend can use this package and maybe show an example, etc

packages/web-data/src/hooks/useIsChainType.ts
```diff
@@ -0,0 +1,15 @@
+import { useEVMContext } from '../context';
+
+export function useIsChainType() {
```
yruej301, line 3: is it missing `isMantle`?

yruej301 (commented):
left some minor comments, i think we should rename the package to something like `react-client`

packages/web-data/README.md
frankfka: Yeah - this is one thing we should look into a lot more. I added a [task](https://www.notion.so/vertexprotocol/Public-Documentation-for-using-vertex-protocol-react-client-aa226a97ad2547aaa60065149cd96941?pvs=4). For now, this is extracted so we can split out Blitz

packages/web-data/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/vertex-protocol/vertex-sdk",
```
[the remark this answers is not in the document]
frankfka: Going to find & replace all of these

packages/web-data/src/hooks/useIsChainType.ts
frankfka: Yeah, guess we never ended up using this hook :P - will add

yruej301 (approved):
looks great, just one tiny thing to fix but otherwise lgtm!

PR #202, opened 2024-06-06
add VRTX supply queries

yruej301:


packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -814,6 +816,20 @@ export class IndexerBaseClient {
...
 
+  /**
+   * Retrieves VRTX total / circulating supply
+   *
+   * @param params
+   */
+  async getVrtxTokenInfo(
+    params: GetIndexerVrtxTokenInfoParams,
+  ): Promise<GetIndexerVrtxTokenInfoResponse> {
+    const baseUrl = this.opts.url.replace('/v1', '');
```
frankfka, line 827: Should we have some sort of `this.v2Url` for when we add more queries in the future?

We should address types as well but can do that later.

Would we be able to just replace /v1 with /v2?

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -676,3 +677,9 @@ export interface IndexerLeaderboardContest {
 export interface GetIndexerLeaderboardContestsResponse {
   contests: IndexerLeaderboardContest[];
 }
+
+export interface GetIndexerVrtxTokenInfoParams {
+  tokenInfoType: IndexerVrtxTokenInfoType;
+}
+
+export type GetIndexerVrtxTokenInfoResponse = number;
```
frankfka, line 685: Is it a number or a string from backend?

frankfka: Is this in X18? ie is it 10^9+18 or just 10^9 being returned?

packages/indexer-client/src/IndexerBaseClient.ts
yruej301: yea, will do!

packages/indexer-client/src/types/clientTypes.ts
yruej301: it's just a number and it's 10^9

packages/indexer-client/src/IndexerBaseClient.ts
yruej301: addressed!

packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -109,6 +112,7 @@ export class IndexerBaseClient {
   readonly axiosInstance: AxiosInstance;
 
   constructor(opts: IndexerClientOpts) {
+    opts.v2Url = opts.v2Url ? opts.v2Url : opts.url.replace('v1', 'v2');
```
frankfka, line 115: Why would we need the override? Would it be safe to just derive this internally for now?

Not sure we should override opts, I feel like it should be immutable as much as possible.

Could we store this in a readonly variable similar to axiosInstance?

packages/indexer-client/src/types/clientTypes.ts
frankfka: Can we leave a JSDoc comment indicating that this does not include decimals?

packages/indexer-client/src/IndexerBaseClient.ts
yruej301: check now

yruej301:
> Can we leave a JSDoc comment indicating that this does not include decimals?

@frankfka do you mean, that is not x18? it does include decimals, see sample responses

```
circulating supply: 244529235.8614397
total supply: 997999999.9999979
```

frankfka:
Decimals as in token decimals, ig im not sure what exactly to call it, but in FE when something includes decimals, it has additional 0s corresponding to the decimals of the token

Ex 1 USDC with decimals is 10^6

yruej301:
gotcha

PR #203, opened 2024-06-10
feat: add back support for arb rewards

Mario857:
It was removed in the past but we need it for round 2


Mario857:
> I assume this was just a revert of the removal PR?

Yup, just a revert.

PR #204, opened 2024-06-14
Add wallet client

cryptoMavrik:
Exposes the wagmi wallet client via hook.


frankfka (approved):
👍 For now we can duplicate this on the trade app for passing builds

cryptoMavrik:
@frankfka should we add optional params to this that drill through jic?

cryptoMavrik:
> @frankfka should we add optional params to this that drill through jic?

A lot of them are linked to the connector, but there are others that potentially could be useful (`batch`) though there aren't any current usecases.

frankfka:
I think we add as we need, but i looked at this again, we should destructure `data` and only pass the wallet client through, otherwise a lot of other stuff from react query is exposed as well

frankfka:
@cryptoMavrik gonna clean this up

PR #206, opened 2024-06-25
Patch wallet client hook

cryptoMavrik:
The hook is not properly initialize without the other params. This PR adds the other params to the `useWalletClient` hook.


frankfka:
@cryptoMavrik what extra params do you need?

cryptoMavrik:
> @cryptoMavrik what extra params do you need?

Currently we aren't using any of them, but it doesn't work without them. 
image

frankfka (approved):

PR #208, opened 2024-07-01
feat: min deposit rates query

yruej301:


apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -399,6 +402,41 @@ async function fullSanity(context: RunContext) {
...
+  const minDepositRates = await client.getMinDepositRates();
+
+  prettyPrint('Min deposit rates', minDepositRates);
+
+  const subInfo = await client.getSubaccountSummary({
+    subaccountOwner: await signer.getAddress(),
+    subaccountName: 'default',
+  });
+
+  for (const balance of subInfo.balances) {
```
frankfka, line 415: nit: We can use `forEach` (a bit more "TS"-y)

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -399,6 +402,41 @@ async function fullSanity(context: RunContext) {
     subaccountName: 'default',
   });
   prettyPrint('Subaccount info after withdraw collateral', subaccountInfoAtEnd);
+
+  const minDepositRates = await client.getMinDepositRates();
+
+  prettyPrint('Min deposit rates', minDepositRates);
+
+  const subInfo = await client.getSubaccountSummary({
```
frankfka, line 410: To reduce number of calls here, could we just use `subaccountInfoAtEnd`?

packages/engine-client/src/utils/queryDataMappers.ts
```diff
@@ -255,3 +259,25 @@ export function mapEngineMarketPrice(
...
+    baseResponse.min_deposit_rates,
+    mapEngineMinDepositRate,
+  );
+
+  return {
+    minDepositRates,
+  };
+}
+
+export function mapEngineMinDepositRate(
```
frankfka, line 276: nit: we might be able to just do this inline in `EngineQueryClient`, these fns aren't used / relevant anywhere else

packages/engine-client/src/utils/queryDataMappers.ts
```diff
@@ -255,3 +259,25 @@ export function mapEngineMarketPrice(
...
+export function mapEngineServerDepositRates(
+  baseResponse: EngineServerMinDepositRatesResponse,
+): GetEngineMinDepositRatesResponse {
+  const minDepositRates: Record<number, EngineMinDepositRate> = mapValues(
+    baseResponse.min_deposit_rates,
+    mapEngineMinDepositRate,
+  );
+
+  return {
+    minDepositRates,
```
frankfka, line 272: Backend nests in an object, but do we need to do this? or could we just return the `Record<number, EngineMinDepositRate>`?

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -399,6 +402,41 @@ async function fullSanity(context: RunContext) {
...
+    if (balance.type == ProductEngineType.SPOT) {
+      const minDepositRate =
+        minDepositRates.minDepositRates[balance.productId].minDepositRate;
+      console.log('product', balance.productId);
+      console.log('min Deposit Rate', minDepositRate.toNumber());
+      console.log(
+        'deposit APR',
+        calcRealizedDepositRateForTimeRange(
+          balance,
+          31536000,
```
frankfka, line 425: Could we use `TimeInSeconds`

yruej301: sure

packages/engine-client/src/utils/queryDataMappers.ts
yruej301: gotcha, sg will clean it up

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -399,6 +407,36 @@ async function fullSanity(context: RunContext) {
...
+
+  const minDepositRates = await client.getMinDepositRates();
+
+  prettyPrint('Min deposit rates', minDepositRates);
+
+  subaccountInfoAtEnd.balances.forEach((balance) => {
+    if (balance.type == ProductEngineType.SPOT) {
+      const minDepositRate =
+        minDepositRates.minDepositRates[balance.productId].minDepositRate;
+      console.log('product', balance.productId);
```
frankfka, line 419: Should we just combine all of these calls? `console.log` takes varargs. You could also just print out an object (can use `toPrintableObject`) - could leave rates as fractions to make life easier too

packages/contracts/src/utils/interest.ts
```diff
@@ -121,12 +122,14 @@ export function calcRealizedDepositRateForTimeRange(
...
 ) {
   const utilization = calcUtilizationRatio(product);
   if (utilization.eq(0)) {
     return toBigDecimal(0);
   }
   return utilization
-    .times(calcBorrowRateForTimeRange(product, seconds))
-    .times(BigDecimals.ONE.minus(toBigDecimal(interestFeeFrac)));
+    .times(calcBorrowRateForTimeRange(product, seconds, toBigDecimal(0)))
+    .times(BigDecimals.ONE.minus(toBigDecimal(interestFeeFrac)))
```
frankfka, line 133: Actually think you might be able to skip these `toBigDecimal` calls

frankfka (approved):
Lets hold off on merging until the FE work is complete

packages/engine-client/src/utils/queryDataMappers.ts
frankfka: We are a bit inconsistent with this unfortunately :( ex. `getAllMarkets` just returns `markets` directly

frankfka: Looking at this more let's just return `minDepositRates` directly, we can switch the outliers over some other time

PR #209, opened 2024-07-03
Sei testnet support

yruej301:


frankfka (approved):
Let's hold off on merging this one as well, will publish the min deposit rate first

PR #216, opened 2024-07-15
add support for fast withdrawal signature

clarkohw:


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -717,3 +719,9 @@ export interface GetIndexerVrtxTokenInfoParams {
  * For instance, 10 VRTX will be represented as 10.0, not 10*10^18.
  */
 export type GetIndexerVrtxTokenInfoResponse = number;
+
+export type GetIndexerFastWithdrawalSignatureParams =
+  IndexerServerFastWithdrawalSignatureParams;
+
+export type GetIndexerFastWithdrawalSignatureResponse =
```
frankfka, line 726: Might be a bit annoying but for clealiness, can we rename `tx_bytes` to camelcase when we convert from server -> client response?

packages/indexer-client/src/types/serverTypes.ts
```diff
@@ -193,6 +194,13 @@ export interface IndexerServerLeaderboardContestsParams {
   contest_ids: number[];
 }
 
+/**
+ * @property {number | string} idx - The submission index of the WithdrawCollateral transaction to be used for fast withdraw.
```
frankfka, line 198: Btw - JS docs can be at the field level (i.e. a comment  right above L201), which I think works better with doc gen

frankfka (commented):
Lgtm, will wait to approve until ABIs and typescript types are updated

packages/indexer-client/src/types/VertexTx.ts
```diff
@@ -26,8 +26,20 @@ export interface VertexLiquidateSubaccountTx {
   };
 }
 
+export interface EndpointWithdrawCollateralTx {
```
frankfka, line 29: Do we need this extra type? In places where we reference this, we could potentially just do `VertexWithdrawCollateralTx['withdraw_collateral']`

PR #222, opened 2024-08-11
Exporting PRIMARY_CHAINS for use in Zod settings schema.

jfywang:


frankfka:
We could add `export const PRIMARY_CHAIN_IDS = PRIMARY_CHAINS.map((chain) => chain.id);` as well

jfywang:
Does this make `PrimaryChainID` redundant? Keeping `PrimaryChainID` for now so that current code which imports it doesn't break.

frankfka:
It does not - `PrimaryChainID` is a type, `PRIMARY_CHAIN_IDS` is an array

frankfka (approved):

PR #224, opened 2024-08-13
Upgrade eslint

comunidadio:
This PR upgrades prettier and migrates to eslint 9.

The recommended set rules have changed slightly, so this PR also fixes newly triggered (minor) violations.
In our case, recommended rule set had some rules set to "warn" in earlier versions.instead of "error" now (no-unused-vars and no-explicit-any)


eslint.config.mjs
frankfka, line 1: Does this need to be `.mjs`? The ESLint docs has this just as `.js`

eslint.config.mjs
```diff
@@ -0,0 +1,54 @@
...
+      '**/dist/**/*',
+      '**/node_modules/**/*',
+    ],
+    languageOptions: {
+      parser: tsParser,
+      parserOptions: {
+        ecmaFeatures: { jsx: true },
+      },
+      globals: {
+        /* replacement for env.node:true in eslintrc */
```
frankfka, line 29: Did we actually need this?

packages/client/src/apis/perp/PerpExecuteAPI.ts
```diff
@@ -1,8 +1,7 @@
 import { BaseVertexAPI } from '../base';
-import { SettlePnlParams } from '@vertex-protocol/contracts';
 
 export class PerpExecuteAPI extends BaseVertexAPI {
-  async settlePnl(params: SettlePnlParams) {
+  async settlePnl(/*params: SettlePnlParams*/) {
```
frankfka, line 4: nit: We could potentially remove the param for now

eslint.config.mjs
```diff
@@ -0,0 +1,54 @@
+import globals from "globals";
+import js from '@eslint/js';
+import tsParser from '@typescript-eslint/parser';
+import tsPlugin from '@typescript-eslint/eslint-plugin';
+import reactPlugin from 'eslint-plugin-react';
+import reactHooksPlugin from 'eslint-plugin-react-hooks';
```
frankfka, line 6: Does this support eslint v9? https://github.com/facebook/react/issues/28313#issue-2132424076

eslint.config.mjs
```diff
@@ -0,0 +1,54 @@
+import globals from "globals";
+import js from '@eslint/js';
+import tsParser from '@typescript-eslint/parser';
```
frankfka, line 3: Out of curiosity, `typescript-eslint` recommends using their own `.config` fn, would this be useful? https://typescript-eslint.io/getting-started

eslint.config.mjs
```diff
@@ -0,0 +1,54 @@
...
+    },
+    plugins: {
+      '@typescript-eslint': tsPlugin,
+      'react': reactPlugin,
+      'react-hooks': reactHooksPlugin,
+      'prettier': prettierPlugin,
+    },
+    rules: {
+      ...js.configs.recommended.rules,
+      ...tsPlugin.configs['eslint-recommended'].rules,
```
frankfka, line 42: Do you know the difference between this & `.recommended` below?

frankfka (commented):
Nice work, leaving a few questions for my own understanding

eslint.config.mjs
frankfka: image

We could hold off on this until all dependencies support v9 - not a huge deal. We could potentially upgrade prettier though

frankfka: I think we can probably park this until a stable release is out

PR #227, opened 2024-08-19
feat: transfer quote

yruej301:


yruej301:
@frankfka it's already on testnet, will make it to mainnet this Thursday

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -437,6 +437,19 @@ async function fullSanity(context: RunContext) {
       );
     }
   });
+
+  const transferQuoteResult = await client.transferQuote({
+    chainId,
+    recipient: subaccountToHex({
```
frankfka, line 443: Should we add some sort of abstraction such that `recipient: Subaccount` to remove the need for consumers to call `subaccountToHex`? It is a bit unclear from the `recipient` naming what format the field expects

packages/engine-client/src/EngineExecuteClient.ts
```diff
@@ -96,4 +96,13 @@ export class EngineExecuteClient extends EngineBaseClient {
       await this.payloadBuilder.buildLinkSignerPayload(params),
     );
   }
+
+  async transferQuote(
```
frankfka, line 100: Should we document (either here or at the top-level client) the restrictions for this? (ex must be to same address, etc.)

apps/e2e/src/engine-client/fullSanity.ts
frankfka: Alternatively, given that we only allow transfers to the same address, potentially just need to pass in the destination subaccount name?

frankfka (commented):
Nice, leaving a few comments

apps/e2e/src/engine-client/fullSanity.ts
yruej301: Yea, I like this.

yruej301: addressed!

packages/contracts/src/eip712/signatureParamTypes.ts
```diff
@@ -73,3 +73,10 @@ export interface EIP712LinkSignerParams extends Subaccount {
   signer: string;
   nonce: string;
 }
+
+export interface EIP712TransferQuoteParams extends Subaccount {
```
frankfka, line 77: So the EIP params should mirror that of what the signature / contract expects, but we can add this "recipientName" abstraction at the engineClient or top-level client

Thoughts on `toSubaccountName` or `recipientSubaccountName` as the property name?

yruej301: i think `recipientSubaccountName` looks good

yruej301: oh i took this pattern from what was done for `EIP712LiquidateSubaccountParams` seemed cleanest to me?

frankfka: Ah that's true 🤔 okay 👍🏻

frankfka (approved):

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -437,6 +437,16 @@ async function fullSanity(context: RunContext) {
       );
     }
   });
+
+  const transferQuoteResult = await client.transferQuote({
+    chainId,
+    recipientSubaccountName: 'transfer1',
+    subaccountOwner: signer.address,
+    subaccountName: 'default',
+    amount: toFixedPoint(50), // amount must be x18
```
tmns, line 446: Does this mean the decimals must be `18`? 

Because when trying with USDC (`6`) I get an error that the amount is below the min (which it shouldn't be). If I add `18` decimals to it though the transfer succeeds with the correct amount. Should I just always be adding `18`?

frankfka: Yeah, decimals should always be 18 (so you would just call `addDecimals` without the second argument)

yruej301:
dis on mainnet now

PR #228, opened 2024-08-21
Add maxTimestampInclusive to Archive Subaccount Pagination Queries

frankfka:
Propagates `maxTimestampInclusive` through subaccount pagination queries. We need this for export history work on the FE


yruej301 (approved):
LGTM

PR #230, opened 2024-08-29
feat: leaderboard updates

yruej301:


yruej301:
changes already on testnet, if all looks good i can push to mainnet

packages/indexer-client/src/dataMappers.ts
```diff
@@ -303,10 +305,22 @@ export function mapIndexerLeaderboardPosition(
...
     accountValue: toBigDecimal(position.account_value),
+    volume: position.volume ? toBigDecimal(position.volume) : undefined,
     updateTime: toBigDecimal(position.update_time),
   };
 }
 
+export function mapIndexerTieredLeaderboardPosition(
+  tieredPosition: IndexerServerTieredLeaderboardPosition,
+): IndexerTieredLeaderboardParticipant {
+  return Object.fromEntries(
```
frankfka, line 316: We could just use `mapValues` inline [here](https://github.com/vertex-protocol/vertex-typescript-sdk/pull/230/files#diff-1a59e593d5688c43202632d4464fd753b47ef4b4d3abf2e0593265a8877f146fR838)

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -682,13 +685,18 @@ export interface GetIndexerLeaderboardResponse {
 }
 
 export interface GetIndexerLeaderboardParticipantParams {
-  contestId: number;
+  contestIds: number[];
   subaccount: Subaccount;
 }
 
+// contestId -> IndexerLeaderboardParticipant;
+export type IndexerTieredLeaderboardParticipant = Record<
```
frankfka, line 693: Hm, we don't necessarily need to be using this for tiered leaderboards (we can use it across unrelated contests, right?), I'm not sure we need this type, as we could define the `Record` directly inline.

Related question - if a subaccount is not eligible, would the key be omitted from the response? If so, this might be good to shout-out in the JSDoc comment

packages/indexer-client/src/dataMappers.ts
yruej301: oh ic ic sg

packages/indexer-client/src/types/clientTypes.ts
yruej301: yup, agree -- will make the changes

frankfka (approved):

yruej301:
this is on mainnet btw

PR #233, opened 2024-09-17
Feat/Blitz Points Leaderboard

Mario857:
Support blitz points leaderboard and add rank to existing blitz points.

### DIFF
--- a/packages/indexer-client/src/types/serverTypes.ts
+++ b/packages/indexer-client/src/types/serverTypes.ts
@@ -168,6 +168,12 @@ export interface IndexerServerBlastPointsParams {
   address: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/indexer-client/src/types/serverTypes.ts
```diff
@@ -168,6 +168,12 @@ export interface IndexerServerBlastPointsParams {
   address: string;
 }
 
+export interface IndexerServerBlitzPointsLeaderboardParams {
+  epoch: number;
+  start?: number | string;
```
frankfka, line 173: nit: can we just enforce that this be a number?

frankfka (approved):

frankfka:
Published `v0.50.0`

PR #234, opened 2024-09-19
feat (blitz points leaderboard): add pagination endpoint

Mario857:
Adds missing pagination endpoint

### DIFF
--- a/apps/e2e/src/indexer-client/fullSanity.ts
+++ b/apps/e2e/src/indexer-client/fullSanity.ts
@@ -97,7 +97,7 @@ async function fullSanity(context: RunContext) {
 
     const blitzPointsLeaderboard =
       await client.getPaginatedBlitzPointsLeaderboard({
-        startCursor: undefined,
+        startCursor: '2',
         epoch: 1,
         limit: 10,
       });
@@ -139,7 +139,7 @@ async function fullSanity(context: RunContext) {
 
   const now = nowInSeconds();
   const multiTimestampProductSnapshots = await client.getMultiProductSnapshots({
-    productIds: [1, 2, 3],
+    productIds: [0, 2, 4],
     maxTimestampInclusive: [
       now,
       now - TimeInSeconds.HOUR,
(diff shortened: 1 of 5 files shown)


packages/indexer-client/src/types/paginatedEventsTypes.ts
```diff
@@ -210,6 +212,13 @@ export type GetIndexerPaginatedLeaderboardParams = IndexerPaginationParams &
 export type GetIndexerPaginatedLeaderboardResponse = WithPaginationMeta &
   GetIndexerLeaderboardResponse;
 
+export type GetIndexerPaginatedBlitzPointsLeaderboardParams =
+  IndexerPaginationParams &
+    Pick<GetIndexerBlitzPointsLeaderboardParams, 'epoch'>;
```
frankfka, line 217: Otherwise we have overridden properties of `limit` & `startCursor`

packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -781,7 +781,7 @@ export class IndexerBaseClient {
   ): Promise<GetIndexerBlitzPointsLeaderboardResponse> {
     const baseResponse = await this.query('blitz_points_leaderboard', {
       limit: params.limit,
-      start: params.startCursor,
+      start: Number(params.startCursor),
```
frankfka, line 784: This was needed, else server throws an error (not sure if you've tested via the sanity test)

apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -138,7 +139,7 @@ async function fullSanity(context: RunContext) {
 
   const now = nowInSeconds();
   const multiTimestampProductSnapshots = await client.getMultiProductSnapshots({
-    productIds: [1, 2, 3],
+    productIds: [0, 2, 4],
```
frankfka, line 142: This was needed to make the sanity test work on blast testnet (product ID of 1 doesn't exist)

PR #236, opened 2024-10-09
feat add staking v2 testnet

cryptoMavrik:
This PR adds the staking v2 abi and necessary execution.


packages/client/src/apis/rewards/types.ts
```diff
@@ -16,3 +16,9 @@ type AmountOrAllParams =
 export type ClaimLiquidTokensParams = {
   epoch: number;
 } & AmountOrAllParams;
+
+export type MigrateStakingV2Params = {
+  staker: AddressLike;
```
frankfka, line 21: I think we usually just use `string` for these cases

frankfka (approved):

packages/engine-client/src/utils/queryDataMappers.ts
```diff
@@ -11,12 +11,9 @@ import {
 import { fromX18, mapValues, toBigDecimal } from '@vertex-protocol/utils';
 import {
   EngineMarketPrice,
-  EngineMinDepositRate,
```
cryptoMavrik, line 14: These weren't being used

frankfka (approved):

PR #245, opened 2024-11-07
Update staking ABIs

cryptoMavrik:


cryptoMavrik:
> Hmm are we now replacing IStaking with Staking?

Just IStakingV2. Updated

frankfka (approved):

PR #247, opened 2024-11-16
Add subaccount and base tx fields for indexer query responses

frankfka:
These essentially expose the fields necessary for us to tell what `subaccountName` transactions are associated with, allowing us to differentiate between iso/cross actions


yruej301 (approved):
lgtm!

PR #251, opened 2024-11-26
chore: update leaderboard registration

yruej301:


apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -337,31 +337,47 @@ async function fullSanity(context: RunContext) {
...
 
-  prettyPrint(
-    'Update leaderboard registration result',
-    updateLeaderboardRegistrationResult,
-  );
+    prettyPrint(
+      'Update leaderboard registration result',
+      updateLeaderboardRegistrationResult,
+    );
+  } catch (e: any) {
```
frankfka, line 356: nit: I think `unknown` works slightly better here, but you might not be able to retrieve `e?.response?.data`. End of the day, not a huge deal

frankfka (approved):

yruej301:
planning to deploy this on Monday so we can merge this after

yruej301:
@frankfka this is on prod now

PR #254, opened 2024-12-03
feat (indexer client): add oracle_prices, add type guard

Mario857:
adds `oracle_prices` to `market_snapshots`, adds type guard to prevent lint error.


apps/e2e/src/indexer-client/fullSanity.ts
```diff
@@ -353,10 +354,10 @@ async function fullSanity(context: RunContext) {
       'Update leaderboard registration result',
       updateLeaderboardRegistrationResult,
     );
-  } catch (e: any) {
+  } catch (e: unknown) {
     console.log(
       'Failed to update leaderboard registration:',
-      e?.response?.data ?? e,
```
Mario857, line 359: This was giving me a lint error so had to add typeguard and switch to unknown.

frankfka: Given that this is just a testing script, I think we can be hacky with how we log these, could just have a `logResponseError` util fn that casts like the following PR does: https://github.com/vertex-protocol/vertex-typescript-sdk/pull/253/files

frankfka (approved):

apps/e2e/src/indexer-client/fullSanity.ts
frankfka: Added this: https://github.com/vertex-protocol/vertex-typescript-sdk/pull/253/files#diff-e9d83b02bf2353d7065359fef6c057f4b29457985ce4b377785b9e3fc5567497R1

PR #257, opened 2024-12-27
Add sonic points queries

cryptoMavrik:
This PR adds interfaces and queries needed for Sonic points launch

### DIFF
--- a/packages/indexer-client/src/types/clientTypes.ts
+++ b/packages/indexer-client/src/types/clientTypes.ts
@@ -601,15 +601,21 @@ export type GetIndexerClaimFoundationRewardsMerkleProofsResponse =
  */
 
 export interface GetIndexerSonicPointsParams {
+  // Subaccount address
   address: string;
 }
 
 export interface GetIndexerSonicPointsResponse {
   tradingPoints: BigDecimal;
   referralPoints: BigDecimal;
-  totalVolume: BigDecimal;
-  rank: BigDecimal;
+  // Total taker volume in quote asset
+  takerVolume: BigDecimal;
+  // Total maker volume in quote asset
+  makerVolume: BigDecimal;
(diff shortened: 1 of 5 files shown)


packages/indexer-client/src/types/serverTypes.ts
```diff
@@ -190,6 +190,15 @@ export interface IndexerServerBlitzInitialDropConditionsParams {
   address: string;
 }
 
+export interface IndexerServerSonicPointsParams {
+  address: string;
+}
+
+export interface IndexerServerSonicPointsLeaderboardParams {
+  start: number;
```
frankfka, line 198: Are these optional?

frankfka (approved):

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -622,6 +661,9 @@ export interface GetIndexerBlitzPointsResponse {
   tradingPoints: BigDecimal;
   // Total accrued points from referrals for phase 1 & phase 2
   referralPoints: BigDecimal;
+  takerVolumes: BigDecimal;
```
frankfka, line 664: What unit is this in? Would be good to specify in a comment.

Additionally, this shouldn't be plural, should it?

cryptoMavrik: Ah it must be quote, so on FE we need to adjust. It doesn't have to be. Plural just matches the response.

frankfka: We should prob rename the response to be non-plural to make more sense (can potentially ask backend to tweak as well). It would then be consistent with the likes of `IndexerSubaccountRewardsForProduct`

frankfka (approved):
Please ask backend to rename the volumes key for a future update

PR #258, opened 2025-01-04
Fix subaccount name conversion by only stripping trailing zero bytes

frankfka:
I found this bug while working on isolated, but the zero-bytes replacement breaks with something like:

`\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0002\u0000iso`

This is because we blindly remove `\u0000`, but this means that reverse conversion to bytes will not produce the same original result. We are only safe to remove trailing zero bytes.

### DIFF
--- a/packages/contracts/src/utils/bytes32.ts
+++ b/packages/contracts/src/utils/bytes32.ts
@@ -19,24 +19,19 @@ import {
  */
 export function subaccountToBytes32(subaccount: Subaccount): SubaccountBytes32 {
   const address = getBytes(subaccount.subaccountOwner);
-  // If the subaccount name is a hex string, then assume that it's a hexlified representation of a non-utf8 string
-  let name: Uint8Array;
-  if (isHexString(subaccount.subaccountName)) {
-    name = getBytes(subaccount.subaccountName);
-  } else {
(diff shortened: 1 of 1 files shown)


yruej301 (approved):
gotcha, changes seem fine but would be good to write some tests for this

packages/contracts/src/utils/bytes32.ts
```diff
@@ -107,20 +114,13 @@ export function subaccountFromHex(subaccount: string): Subaccount {
   return subaccountFromBytes32(getBytes(subaccount));
 }
 
-export function strToBytes(input: string, bytesLen: number): Bytes {
```
frankfka, line 110: Decided to combine this into `subaccountNameToBytes12` as we have no other usage of it. We can then use `subaccountNameToBytes12` in `subaccountToBytes32` so that we can share the custom `isHexString` logic

frankfka:
We're long overdue for some SDK tests, its getting to the point where its a bit scary to make changes in foundational files. Going to add to the board

yruej301:
> We're long overdue for some SDK tests, its getting to the point where its a bit scary to make changes in foundational files. Going to add to the board

sounds good! and agree

PR #260, opened 2025-01-08
feat (engine client): add insurance.

Mario857:
Adds `getInsurance` to engine client.


packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -528,6 +529,16 @@ export class EngineQueryClient extends EngineBaseClient {
     };
   }
 
+  /**
+   * Gets the insurance funds
```
frankfka, line 533: In what unit? We should specify in the comment

frankfka (approved):

PR #261, opened 2025-01-09
Add Unit Test Support + `bytes32` unit tests

frankfka:


packages/contracts/src/utils/bytes32.test.ts
```diff
@@ -0,0 +1,112 @@
+import { expect, it } from '@jest/globals';
```
yruej301, line 1: ✨

yruej301 (approved):
this is great!

PR #264, opened 2025-01-23
feat (ethers migration): bigNumberish -> bigDecimalish, bigint

Mario857:


packages/contracts/src/typechain-types/IAirdrop.ts
```diff
@@ -37,15 +36,15 @@ export interface IAirdropInterface extends Interface {
 
   encodeFunctionData(
     functionFragment: "claim",
-    values: [BigNumberish, BigNumberish, BigNumberish, BytesLike[]]
+    values: [bigint, bigint, bigint, BytesLike[]]
```
Mario857, line 39: So I assume we want all these `typechain-types` to use `bigint`

frankfka: No, `typechain` is required only for `ethers` because `ethers` does not have built-in support for Typescript. Once we make the transition, we will be getting rid of `typechain` entirely. `viem` has built-in TS support: https://viem.sh/docs/contract/getContract

These files are autogenerated so should not be updated. We should only remove any `BigNumberish` imports that we reference within our own code

packages/contracts/src/execute/depositCollateral.ts
```diff
@@ -25,8 +25,17 @@ export async function depositCollateral({
...
+      productId,
+      toBigDecimal(amount).toFixed(0),
+      referralCode,
+    );
   } else {
-    return endpoint.depositCollateral(bytesSubaccountName, productId, amount);
+    return endpoint.depositCollateral(
+      bytesSubaccountName,
+      productId,
+      toBigDecimal(amount).toFixed(0),
```
Mario857, line 38: So casting all to `string` instead of previously `bigint`. Because it seems like it can't be serialized.

Mario857:
lmk if this makes sense. I will need to test this a bit more.

packages/trigger-client/src/types/clientTypes.ts
```diff
@@ -44,7 +43,7 @@ export type TriggerOrderStatus =
 interface SignatureParams {
   // Orderbook address for placement, endpoint address for cancellation & listing
   verifyingAddr: string;
-  chainId: BigNumberish;
+  chainId: BigDecimalish;
```
frankfka, line 46: Could this just be a `number`? It seems like `viem`/`wagmi` types `chainId: number`

apps/e2e/src/engine-client/fullSanity.ts
```diff
@@ -116,7 +116,7 @@ async function fullSanity(context: RunContext) {
     subaccountName: 'default',
     amount: toFixedPoint(-0.03),
     expiration: getExpiration(),
-    price: 110000,
```
frankfka, line 119: If inputs are all `BigDecimalish`, doesn't a `number` work as well?

packages/client/src/apis/base.ts
```diff
@@ -2,9 +2,9 @@ import {
   getChainIdFromSigner,
   WithContracts,
 } from '@vertex-protocol/contracts';
-import { BigNumberish } from 'ethers/lib.esm';
 import { VertexClientContext } from '../context';
 import { isSigner } from '../utils';
+import { BigNumberish } from 'ethers';
```
frankfka, line 7: Should this be addressed?

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -214,6 +227,11 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
       return availableAmount.toFixed(0);
     })();
 
-    return [params.epoch, amountToClaim, totalAmount.toFixed(0), proof];
+    return [
+      toBigDecimal(params.epoch).toFixed(),
```
frankfka, line 231: `epoch` is a number, right? why do we need `toFixed`?

packages/contracts/src/eip712/getVertexEIP712Values.ts
```diff
@@ -111,8 +111,8 @@ function getWithdrawCollateralValues(
       subaccountName: params.subaccountName,
     }),
     productId: params.productId,
-    amount: BigInt(params.amount).toString(),
-    nonce: BigInt(params.nonce).toString(),
+    amount: toBigDecimal(params.amount).toFixed(0),
+    nonce: toBigDecimal(params.nonce).toFixed(0),
```
frankfka, line 115: How come this has `toBigDecimal` but others were removed?

packages/contracts/src/eip712/getVertexEIP712Values.ts
```diff
@@ -160,15 +160,15 @@ function getIsolatedOrderValues(
...
+    margin: toBigDecimal(params.margin).toFixed(0),
   };
 }
 
 function getListTriggerOrdersValues(
   params: EIP712ListTriggerOrdersParams,
 ): EIP712ListTriggerOrdersValues {
   return {
-    recvTime: BigInt(params.recvTime).toString(),
+    recvTime: params.recvTime,
```
frankfka, line 171: How come this doesn't need string conversion anymore, when it's a `BigDecimalish`?

packages/contracts/src/execute/depositCollateral.ts
frankfka: Hmm that's very weird, this is going through ethers, where `BigNumberish` includes `bigint`. 

I don't see any related issues: https://github.com/ethers-io/ethers.js/issues?q=is%3Aissue%20Do%20not%20know%20how%20to%20serialize%20a%20BigInt 

Perhaps this is worth opening an issue with them?

frankfka: Also might be worth extracting a few utils (ex. `toBigInt(val: BigDecimalish)`, `toIntegerString(val: BigDecimalish)`)

Mario857: So my bad this can stay `bigint` and use `toBigInt` as we are calling `ethers` directly.

I realized there is a [comment](https://github.com/vertex-protocol/vertex-sdk/blob/9a3b1655b39ae1596d00f692ece97d540f156b5a/packages/contracts/src/eip712/getVertexEIP712Values.ts#L40) about it for `getVertexEIP712Values` where we have json stringiny so this has to use `toNumberString`.

packages/contracts/src/execute/depositCollateral.ts
```diff
@@ -25,8 +29,12 @@ export async function depositCollateral({
   if (referralCode) {
     return endpoint[
       'depositCollateralWithReferral(bytes12,uint32,uint128,string)'
-    ](bytesSubaccountName, productId, amount, referralCode);
+    ](bytesSubaccountName, productId, toIntegerString(amount), referralCode);
```
frankfka, line 32: Why is this using `toIntegerString`, but L37 uses `toBigInt`?

frankfka (commented):
Things look good! i'll pull this locally and use the local SDK in my web dev for the next few days, just to give more opportunity for issues to arise

packages/contracts/src/execute/depositCollateral.ts
Mario857: Both should use toBigInt actually.

frankfka (approved):

PR #265, opened 2025-01-26
Explicitly specify 0dp in BigDecimal.toFixed()

frankfka:
This was an oversight on my part. We use `toFixed()` to convert a `BigDecimal` to a `string` with 0 decimal places. Omitting the `0` param in `Number.toFixed()` defaults to 0dp, but this isn't the case for `BigDecimal` (https://mikemcl.github.io/bignumber.js/#toFix):

image


yruej301 (approved):
just seeing this! looks good -- thanks for the fix!

PR #268, opened 2025-02-07
feat (isolated): add isolated and isolatedProductId

Mario857:


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -473,6 +473,9 @@ export interface IndexerProductPayment {
   // Represents the annually interest rate for spots and annually funding rate for perps.
   annualPaymentRate: BigDecimal;
   oraclePrice: BigDecimal;
+  isolated: boolean;
+  // Only defined when primaryQuoteProduct (productId=0) and isolated=true otherwise null.
```
frankfka, line 477: Can we reuse the same comment as we have currently in `serverModelTypes`?

`// The product ID associated with the isolated perp market. This is only used when product_id === QUOTE_PRODUCT_ID and isolated === true`

PR #272, opened 2025-02-17
chore: avax testnet

yruej301:


frankfka (approved):
Do we have any spot markets on avax?

yruej301:
> Do we have any spot markets on avax?

not yet

PR #276, opened 2025-02-26
Migrate entirety of SDK to use Viem

frankfka:


packages/contracts/src/common/abis/Airdrop.ts
frankfka, line 1: ABIs now need to be in typescript for type inference to work :(

packages/client/src/createVertexClient.ts
```diff
@@ -14,7 +14,7 @@ import {
  */
 export function createVertexClient(
   opts: CreateVertexClientContextOpts,
-  signerOpts: CreateVertexClientContextSignerOpts,
+  signerOpts: CreateVertexClientContextAccountOpts,
```
Mario857, line 17: nit: `accountOpts`

apps/e2e/src/utils/runWithContext.ts
```diff
@@ -1,25 +1,39 @@
...
+      account,
+      chain: CHAIN_ENV_TO_CHAIN[env.chainEnv],
+      transport: http(),
+    });
   };
 
+  const publicClient = createPublicClient({
+    chain: CHAIN_ENV_TO_CHAIN[env.chainEnv],
+    transport: http(),
+    // The cast below is needed for some reason
```
Mario857, line 30: I think it's because of the chain type. Which is a bit different for each chain.

Mario857 (commented):
In general looking good. Do all sanity test pass on your side?

apps/e2e/src/utils/runWithContext.ts
frankfka: Perhaps, yeah

frankfka:
> In general, looking good. Do all sanity tests pass on your side?

Yep, sanity tests do pass 👍🏻

apps/e2e/src/client/fullSanity.ts
```diff
@@ -4,86 +4,98 @@ import {
...
+      productId: 0,
+      amount: toFixedPoint(10000, 6),
+    }),
+    publicClient,
+  );
+
+  const allMarkets = await vertexClient.market.getAllMarkets();
 
   console.log('Placing order...');
+  const spotOrderProductId = 4;
```
cryptoMavrik, line 81: Wouldn't this be a perp market?

frankfka: Ah yeah, this should be `3`

cryptoMavrik (approved):

Mario857 (approved):

PR #277, opened 2025-03-06
chore: sunset bera-test

yruej301:


yruej301:
@frankfka `bera-test` has been sunset, will prob launch a new bera testnet next week under `bepolia-test`

frankfka:
@yruej301 Can we replace `beraTestnet` with the new config when deployed? Otherwise any new package versions will break typing in web

yruej301:
> @yruej301 Can we replace `beraTestnet` with the new config when deployed? Otherwise any new package versions will break typing in web

yup, we can do that -- will follow up in a new PR with the changes next week

PR #278, opened 2025-03-07
Vertex SDK V1

frankfka:


frankfka:
Will run `typedoc` after this is merged

PR #279, opened 2025-03-10
Feat/staking v2 queries

Mario857:


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -899,3 +899,35 @@ export interface GetIndexerFastWithdrawalSignatureResponse {
...
+}
+
+export interface GetIndexerStakingV2PoolSnapshotsParams {
+  // Currently accepts all integers, in seconds
+  granularity: number;
+  // Seconds
+  maxTimeInclusive?: number;
+  limit: number;
+}
+export interface GetIndexerStakingV2PoolSnapshotsResponse {
```
frankfka, line 917: nit: space above

packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -1087,6 +1093,44 @@ export class IndexerBaseClient {
...
+   *
+   * @param params
+   */
+  async getStakingV2PoolSnapshots(
+    params: GetIndexerStakingV2PoolSnapshotsParams,
+  ): Promise<GetIndexerStakingV2PoolSnapshotsResponse> {
+    const baseResponse = await this.query('staking_v2_pool_snapshots', {
+      interval: {
+        count: params.limit,
+        max_time: params.maxTimeInclusive?.toString(),
```
frankfka, line 1124: nit: would a `toFixed` be a better conversion here?

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -899,3 +899,35 @@ export interface GetIndexerFastWithdrawalSignatureResponse {
...
+  snapshots: IndexerStakingV2PoolSnapshot[];
+}
+
+export interface IndexerStakingV2Staker {
+  address: Address;
+  stakedAmount: BigDecimal;
+  poolShare: BigDecimal;
```
frankfka, line 924: Could we add a JSDoc to explain that this is a fraction?

Also, for smaller numbers, we prob don't need `BigDecimal`. However, this does make it more consistent with other fields, so i'll leave the decision up to you

packages/indexer-client/src/types/serverModelTypes.ts
```diff
@@ -178,6 +178,19 @@ export interface IndexerServerMarketSnapshot {
...
+  timestamp: string;
+  cumulative_staked: string;
+  cumulative_unstaked: string;
+  number_of_stakers: string;
+}
+
+export interface IndexerServerStakingV2Staker {
+  address: string;
+  stake_amount: string;
+  pool_share: string;
```
frankfka, line 191: I'm surprised it isn't a `number` that Chang returns here :o

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -899,3 +899,35 @@ export interface GetIndexerFastWithdrawalSignatureResponse {
...
+}
+
+export interface GetIndexerStakingV2TopStakersParams {
+  limit: number;
+}
+
+export type GetIndexerStakingV2TopStakersResponse = {
```
frankfka, line 931: nit: `interface`

packages/indexer-client/src/types/serverModelTypes.ts
Mario857: It is number actually, my bad.

packages/indexer-client/src/IndexerBaseClient.ts
Mario857: Will switch all to toFixed().

frankfka: Should this be `toFixed(0)`?

Mario857: I don't think there is a difference.

frankfka: On mobile so I cannot check, but if this is a BigDecimal then there is: https://mikemcl.github.io/bignumber.js/#toFix

Mario857: https://github.com/vertex-protocol/vertex-typescript-sdk/pull/280/files

PR #283, opened 2025-03-25
Add VLP executes and queries

frankfka:


frankfka:
@yruej301 Bumping this

packages/engine-client/src/EngineQueryClient.ts
```diff
@@ -531,6 +533,29 @@ export class EngineQueryClient extends EngineBaseClient {
     };
   }
 
+  /**
+   * Retrieves the estimated max base * quote amounts for minting VLP. Only the quote amount is used for minting.
```
yruej301, line 537:
```suggestion
   * Retrieves the estimated max base & quote amounts for minting VLP. Only the quote amount is used for minting.
```

yruej301 (approved):
looks great, a small typo but other than that -- send it!

yruej301:
@frankfka actually didn't see any tests, might be worth adding some

frankfka:
@yruej301 added tests!

PR #284, opened 2025-03-31
feat (indexer): add vrtx supply and foundation token incentives query

Mario857:
Screenshot 2025-04-08 at 20 54 19

Screenshot 2025-04-08 at 20 54 45

### DIFF
--- a/packages/indexer-client/src/IndexerBaseClient.ts
+++ b/packages/indexer-client/src/IndexerBaseClient.ts
@@ -1174,7 +1174,9 @@ export class IndexerBaseClient {
     );
 
     return {
-      snapshots: baseResponse.snapshots.map(mapIndexerFoundationTokenSnapshot),
+      snapshots: mapValues(baseResponse.snapshots, (snapshots) =>
+        snapshots.map(mapIndexerFoundationTokenSnapshot),
+      ),
     };
   }
 
--- a/packages/indexer-client/src/IndexerBaseClient.ts
+++ b/packages/indexer-client/src/IndexerBaseClient.ts
@@ -26,6 +26,7 @@ import {
   mapIndexerEvent,
   mapIndexerEventWithTx,
   mapIndexerFoundationTakerRewardsWeek,
+  mapIndexerFoundationTokenSnapshot,
   mapIndexerFundingRate,
   mapIndexerLeaderboardContest,
   mapIndexerLeaderboardPosition,
@@ -67,6 +68,7 @@ import {
   GetIndexerFastWithdrawalSignatureResponse,
   GetIndexerFoundationTakerRewardsParams,
   GetIndexerFoundationTakerRewardsResponse,
+  GetIndexerFoundationTokenIncentivesSnapshotsParams,
   GetIndexerFundingRateParams,
   GetIndexerFundingRateResponse,
   GetIndexerInterestFundingPaymentsParams,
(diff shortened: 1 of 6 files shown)


packages/indexer-client/src/types/serverModelTypes.ts
```diff
@@ -349,3 +364,10 @@ export interface IndexerServerStakingV2PoolSnapshotsInterval {
   granularity: number;
   max_time?: string;
 }
+
+export interface IndexerServerVrtxSupplySnapshotsInterval {
```
frankfka, line 368: Is there a common type that can be extracted for these interval interfaces?

frankfka, line 368: Is there a common type that can be extracted for these interval interfaces?

packages/indexer-client/src/types/serverModelTypes.ts
```diff
@@ -191,6 +195,17 @@ export interface IndexerServerStakingV2Staker {
...
 }
 
+export interface IndexerServerVrtxSupplySnapshot {
+  timestamp: string;
+  vrtx_oracle_price: string;
+  cumulative_incentives: string;
+  cumulative_lba: string;
+  cumulative_supply_ecosystem: string;
+  cumulative_supply_treasury: string;
+  cumulative_investors_supply: string;
```
frankfka, line 205: Should we get backend to change the names such that the placement of `supply` in the naming is consistent? There are two fields with supply as a prefix and two with supply as a postfix

packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -1131,6 +1133,20 @@ export class IndexerBaseClient {
     };
   }
 
+  async getVrtxSupplySnapshots(params: GetIndexerVrtxSupplySnapshotsParams) {
+    const baseResponse = await this.query('vrtx_supply_snapshots', {
+      interval: {
+        count: params.limit,
+        max_time: params.maxTimeInclusive?.toFixed(0),
```
frankfka, line 1140: `toIntegerString` may be useful for some of these

packages/indexer-client/src/IndexerBaseClient.ts
```diff
@@ -1137,6 +1141,45 @@ export class IndexerBaseClient {
     };
   }
 
+  async getVrtxSupplySnapshots(params: GetIndexerVrtxSupplySnapshotsParams) {
+    const baseResponse = await this.query('vrtx_supply_snapshots', {
+      interval: {
+        count: params.limit,
```
frankfka, line 1147: Could we add a follow-up task to extract this logic into a util fn?

packages/indexer-client/src/dataMappers.ts
```diff
@@ -434,3 +438,37 @@ export function mapIndexerStakingV2Staker(
     poolShare: staker.pool_share,
   };
 }
+
+export function mapIndexerVrtxSupplySnapshot(
+  snapshot: IndexerServerVrtxSupplySnapshot,
+): IndexerVrtxSupplySnapshot {
+  return {
+    timestamp: toBigDecimal(snapshot.timestamp),
+    vrtxOraclePrice: toBigDecimal(snapshot.vrtx_oracle_price),
```
frankfka, line 447: should this be `fromX18(...)` ?

packages/indexer-client/src/dataMappers.ts
```diff
@@ -434,3 +438,37 @@ export function mapIndexerStakingV2Staker(
...
+
+export function mapIndexerFoundationTokenIncentivesSnapshot(
+  snapshot: IndexerServerFoundationTokenIncentivesSnapshot,
+): IndexerFoundationTokenIncentivesSnapshot {
+  return {
+    timestamp: toBigDecimal(snapshot.timestamp),
+    cumulativeFoundationTokenIncentives: toBigDecimal(
+      snapshot.cumulative_foundation_token_incentives,
+    ),
+    foundationTokenOraclePrice: toBigDecimal(
```
frankfka, line 469: Same thing here

packages/indexer-client/src/IndexerBaseClient.ts
frankfka: Or if it's not a big change, we can do that in this PR

frankfka (approved):

packages/indexer-client/src/IndexerBaseClient.ts
frankfka: Could be something like `mapSnapshotsIntervalToServerParams`?

PR #286, opened 2025-04-08
Do not omit maxTimestampInclusive from GetIndexerSubaccountInterestFundingPaymentsParams

comunidadio:
Context: #3135 in monorepo

This PR removes omit for maxTimestampInclusive as it is now supported on backend.


packages/utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vertex-protocol/utils",
-  "version": "1.2.1",
+  "version": "1.2.2",
```
frankfka, line 3: did you somehow publish a version?

frankfka: Ah no this is automatically handled by lerna

frankfka (approved):

PR #288, opened 2025-04-09
fix (utils): remove fixedPoint

Mario857:


packages/utils/src/math/decimalAdjustment.ts
```diff
@@ -50,3 +51,21 @@ export function removeDecimals<T extends number | BigDecimal | undefined>(
 
   return getResult() as T;
 }
+
+/**
+ * Converts a value to X18 fixed point representation
+ *
+ * @param val
+ */
+export function toX18(val: BigDecimalish): bigint {
```
Mario857, line 60: Does it make sense to replace these toX18 with `toBigInt(addDecimals(toBigDecimal(value))`? I think it's a bit verbose.

Mario857: https://github.com/vertex-protocol/vertex-typescript-sdk/pull/288/commits/20386414fb8bbca416fe7fe7d573bd2f25582530

https://github.com/vertex-protocol/vertex-typescript-sdk/pull/288/commits/71024d33c60fe00af6eafc2e75013012ebfd473c

frankfka: Seems like majority of additions isn't the `bigint` requirement but actually converting to `BigDecimal`

What if `add/removeDecimals` allowed `BigDecimalish`?

Mario857: Yeah wanted to do that but typing was a bit tricky. Might need to do typing for function overloads

Like
```
export function addDecimals(value: undefined, decimals?: number): undefined;
export function addDecimals(value: number, decimals?: number): number;
export function addDecimals(
  value: BigDecimalish,
  decimals?: number,
): BigDecimal;

export function addDecimals(
  value: BigDecimalish | undefined,
  decimals: number = VERTEX_PRODUCT_DECIMALS,
): BigDecimalish | undefined {
  if (value == null) {
    return undefined;
  }

  const adjustedValue = toBigDecimal(value).multipliedBy(
    toBigDecimal(10).pow(decimals),
  );
  return typeof value === 'number' ? adjustedValue.toNumber() : adjustedValue;
}
```

frankfka: What about

```tsx
type NumberOutIfNumberIn<T extends BigDecimalish | undefined> =
  T extends undefined ? undefined : T extends number ? number : BigDecimal;

/**
 * Adds the specified # of decimals to the number. For example, value = 1, decimals = 2, returns 100.
 *
 * @param value can be undefined for better developer experience. If undefined, returns undefined.
 * @param decimals number of decimal places to add, defaults to 18, which is the standard within Vertex
 */
export function addDecimals<T extends BigDecimalish | undefined>(
  value: T,
  decimals: number = VERTEX_PRODUCT_DECIMALS,
): NumberOutIfNumberIn<T> {
  const getResult = () => {
    if (value == null) {
      return undefined;
    }

    const adjustedValue = toBigDecimal(value).multipliedBy(
      toBigDecimal(10).pow(decimals),
    );
    return typeof value === 'number' ? adjustedValue.toNumber() : adjustedValue;
  };

  return getResult() as NumberOutIfNumberIn<T>;
}
```

I don't love the naming of `NumberOutIfNumberIn` though.

Ideally we wouldn't even need this `NumberOutIfNumberIn` behavior - does FE require it?

Mario857: Only one occurence https://github.com/vertex-protocol/vertex-web-monorepo/pull/3188/files

packages/utils/src/math/decimalAdjustment.ts
```diff
@@ -1,18 +1,21 @@
-import { BigDecimal, toBigDecimal } from './bigDecimal';
+import { BigDecimal, BigDecimalish, toBigDecimal } from './bigDecimal';
 
 // All Vertex balances have 18 decimals. Ex. 1e18 = 1.0
 export const VERTEX_PRODUCT_DECIMALS = 18;
 
+type AdjustDecimalsResult<T extends BigDecimalish | undefined> =
+  T extends undefined ? undefined : T extends number ? number : BigDecimal;
```
frankfka, line 7: nit: space below. We may also benefit from a comment here

packages/trigger-client/src/dataMappers.ts
```diff
@@ -91,7 +91,7 @@ export function mapServerOrderInfo(
     amount: toBigDecimal(serverOrder.order.amount),
     expiration: toBigDecimal(serverOrder.order.expiration),
     nonce: serverOrder.order.nonce,
-    price: fromX18(serverOrder.order.priceX18),
+    price: removeDecimals(toBigDecimal(serverOrder.order.priceX18)),
```
frankfka, line 94: Is the `toBigDecimal` needed?

packages/indexer-client/src/dataMappers.ts
```diff
@@ -431,7 +437,9 @@ export function mapIndexerMarketSnapshot(
     ),
     cumulativeInflows: mapValues(snapshot.cumulative_inflows, toBigDecimal),
     cumulativeOutflows: mapValues(snapshot.cumulative_outflows, toBigDecimal),
-    oraclePrices: mapValues(snapshot.oracle_prices, fromX18),
+    oraclePrices: mapValues(snapshot.oracle_prices, (value) =>
```
frankfka, line 440: Could this just be `mapValues(snapshot.oracle_prices, removeDecimals)`?

frankfka: Same comment with above

frankfka (approved):
Some nits but otherwise looks good!

packages/indexer-client/src/dataMappers.ts
Mario857: `removeDecimals` hasa  second parameter which is decimals

packages/trigger-client/src/dataMappers.ts
Mario857: Yeah, needs casting because `bigDecimalish` could be either `number | BigNumber`.

PR #292, opened 2025-05-06
feat (e2e tests): refactor engine tests

Mario857:


apps/e2e/package.json
```diff
@@ -14,8 +14,17 @@
...
     "client/ws-message-tests": "ts-node ./src/client/wsMessageTests.ts",
-    "engine/full-sanity": "ts-node ./src/engine-client/fullSanity.ts",
-    "indexer/full-sanity": "ts-node ./src/indexer-client/fullSanity.ts",
+    "client/full-sanity": "run-s client/collateral-tests client/order-tests client/query-tests client/ws-message-tests",
+    "engine/collateral-tests": "ts-node ./src/engine-client/collateralTests.ts",
+    "engine/lp-tests": "ts-node ./src/engine-client/lpTests.ts",
+    "engine/query-tests": "ts-node ./src/engine-client/queryTests.ts",
+    "engine/signer-order-tests": "ts-node ./src/engine-client/signerAndOrderTests.ts",
+    "engine/vlp-tests": "ts-node ./src/engine-client/vlpTests.ts",
+    "engine/full-sanity": "run-s engine/query-tests engine/collateral-tests engine/signer-order-tests engine/lp-tests engine/vlp-tests",
```
Mario857, line 23: I have added `npm-run-all` instead of defining `allTests` file. Because we wouldn't be able to run tests individually without calling `runWithContext` twice.

frankfka: One potential thing you _could_ try is to just have `allTests` import (or require) the individual files, maybe that results in invocation of the functions?

A bit of a weird pattern though...

apps/e2e/src/engine-client/collateralTests.ts
```diff
@@ -0,0 +1,97 @@
...
+  );
+  await waitForTransaction(
+    quote.write.approve([endpointAddr, amount]),
+    publicClient,
+  );
+
+  // Deposit collateral
+  await waitForTransaction(
+    depositCollateral({
+      amount: addDecimals(10000, 6),
```
frankfka, line 52: Could this be `amount`?

apps/e2e/src/engine-client/collateralTests.ts
```diff
@@ -0,0 +1,97 @@
...
+      endpoint,
+      productId: 0,
+      subaccountName: 'default',
+    }),
+    publicClient,
+  );
+  console.log('Done depositing collateral');
+
+  // Wait for slow mode delay
+  await new Promise((resolve) => setTimeout(resolve, 5000));
```
frankfka, line 62: Thoughts on a `delay` util fn?

apps/e2e/src/engine-client/collateralTests.ts
```diff
@@ -0,0 +1,97 @@
...
+  prettyPrint('Done withdrawing collateral, result', withdrawResult);
+
+  const subaccountInfoAtEnd = await client.getSubaccountSummary({
+    subaccountOwner: walletClientAddress,
+    subaccountName: 'default',
+  });
+  prettyPrint('Subaccount info after withdraw collateral', subaccountInfoAtEnd);
+
+  // Delay for rate limit
+  await new Promise((resolve) => setTimeout(resolve, 5000));
```
frankfka, line 93: Why do we need to do this if there are no other actions after?

apps/e2e/src/engine-client/lpTests.ts
```diff
@@ -0,0 +1,59 @@
...
+  const burnSpotLpResult = await client.burnLp({
+    subaccountOwner: walletClientAddress,
+    subaccountName: 'default',
+    productId: 3,
+    amount: addDecimals(1),
+    verifyingAddr: endpointAddr,
+    chainId,
+  });
+  // Delay for rate limit
+  await new Promise((resolve) => setTimeout(resolve, 5000));
```
frankfka, line 54: Same thing here

apps/e2e/src/trigger-client/fullSanity.ts
```diff
@@ -219,4 +219,5 @@ async function fullSanity(context: RunContext) {
   prettyPrint('List orders by digest result', ordersByDigest);
 }
 
-runWithContext(fullSanity);
+console.log('Running full sanity tests');
```
frankfka, line 222: nit: can we standardize log messages into some format where the client being tested (in this case, the `trigger-client`) is clear?

apps/e2e/src/indexer-client/marketQueriesTests.ts
frankfka, line 1: some of these queries (ex. staking, vlp, rewards) seem unrelated to markets?

apps/e2e/src/engine-client/signerAndOrderTests.ts
```diff
@@ -1,130 +1,44 @@
...
 import { privateKeyToAccount } from 'viem/accounts';
 import { getExpiration } from '../utils/getExpiration';
 import { prettyPrint } from '../utils/prettyPrint';
-import { runWithContext } from '../utils/runWithContext';
 import { RunContext } from '../utils/types';
-import { waitForTransaction } from '../utils/waitForTransaction';
+import { runWithContext } from '../utils/runWithContext';
 
-async function fullSanity(context: RunContext) {
+export async function signerAndOrderTests(context: RunContext) {
```
frankfka, line 22: Should tests be assuming that a subaccount exists with collateral? Otherwise, we should be depositing collateral at the beginning of the tests and withdrawing a roughly equivalent amt at the end.

This might introduce a lot of boilerplate though, so we could also have some `setup` script that mints & deposits to create an account

apps/e2e/package.json
Mario857: Yeah, just importing that resulted in the invocation of the function.

apps/e2e/package.json
```diff
@@ -28,5 +37,8 @@
     "dotenv": "^16.3.1",
     "viem": "*"
   },
-  "gitHead": "6d0f498829ddaf01a11d2f46f960811428b370f5"
+  "gitHead": "6d0f498829ddaf01a11d2f46f960811428b370f5",
+  "devDependencies": {
```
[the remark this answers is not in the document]
frankfka: Shoot, apologies, Im not sure what the original comment was referring to, but We should only be sorting `scripts` and `dependencies` objects. It's usually standard to have `name` / `version` / `type` etc at the very top

apps/e2e/src/indexer-client/subaccountQueriesTests.ts
```diff
@@ -0,0 +1,180 @@
...
+  });
+
+  const subaccount: Subaccount = {
+    subaccountName: 'default',
+    subaccountOwner: walletClient.account.address,
+  };
+
+  const summary = await client.getMultiSubaccountSnapshots({
+    subaccounts: [subaccount],
+    timestamps: [nowInSeconds(), nowInSeconds() - 60 * 60 * 24],
```
frankfka, line 23: nit: could this be `TimeInSeconds.DAY`?

apps/e2e/src/indexer-client/leaderboardTests.ts
```diff
@@ -0,0 +1,135 @@
...
+    url: context.endpoints.indexer,
+    walletClient,
+  });
+
+  const subaccount: Subaccount = {
+    subaccountName: 'default',
+    subaccountOwner: walletClient.account.address,
+  };
+
+  if (
```
frankfka, line 23: Should these be part of rewards?

apps/e2e/src/client/accountSetup.ts
```diff
@@ -0,0 +1,50 @@
...
+async function accountSetup(context: RunContext) {
+  const walletClient = context.getWalletClient();
+  const publicClient = context.publicClient;
+
+  const vertexClient: VertexClient = createVertexClient(context.env.chainEnv, {
+    walletClient,
+    publicClient,
+  });
+
+  const quoteProductId = 0;
```
frankfka, line 16: This could just use `QUOTE_PRODUCT_ID`

apps/e2e/src/client/accountSetup.ts
```diff
@@ -0,0 +1,50 @@
...
+  const quoteMintAmount = addDecimals(1000, 6);
+  const quoteDepositAmount = addDecimals(500, 6);
```
frankfka, line 18: Was there a reason that we deposit less than we mint?

frankfka: Ah, is it for `quoteDepositWithReferralAmount`?

Should we make `collateralTests` runnable by itself? i.e. it mints, deposits, and withdraws without requiring this script

apps/e2e/src/client/orderTests.ts
```diff
@@ -146,10 +122,12 @@ async function orderTests(context: RunContext) {
 
   await vertexClient.spot.withdraw({
```
frankfka, line 123: I don't think we need withdrawals here now that we've removed the deposit

apps/e2e/src/indexer-client/rewardsQueriesTests.ts
```diff
@@ -0,0 +1,28 @@
+import { IndexerClient } from '@vertex-protocol/indexer-client';
+import { runWithContext } from '../utils/runWithContext';
+import { RunContext } from '../utils/types';
+import { TimeInSeconds } from '@vertex-protocol/utils';
+import { prettyPrint } from '../utils/prettyPrint';
+
+async function rewardsQueriesTests(context: RunContext) {
```
frankfka, line 7: What about `getRewards`, `getTakerRewards`, `getClaimVrtxMerkleProofs`, `getClaimFoundationRewardsMerkleProofs`, `getSonicPoints`, and `getSonicPointsLeaderboard`?

apps/e2e/src/engine-client/lpTests.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  const endpointAddr = await clearinghouse.read.getEndpoint();
+
+  const mintSpotLpResult = await client.mintLp({
+    subaccountOwner: walletClientAddress,
+    subaccountName: 'default',
+    productId: 3,
+    amountBase: addDecimals(1),
+    quoteAmountLow: addDecimals(1000),
```
frankfka, line 32: Could we change this to use oracle price * 0.5 (for low) & oracle price * 1.5 (for high)?

apps/e2e/src/engine-client/collateralTests.ts
frankfka, line 1: Realized that this is essentially a duplicate of `collateralTests` for the top level client, we might just want to keep one of them around (probably the top level client)

apps/e2e/src/engine-client/accountSetup.ts
frankfka, line 1: I think we just need one account setup script

apps/e2e/package.json
frankfka: Does anything change with [tsx](https://tsx.is/dev-api/entry-point)? 

Now that I look at the `package.json`, I can see these scripts getting a bit out of hand in terms of command length. It might be cleaner to define a `full-sanity` file that just `import`'s sub-tests

Mario857: Also convert `runWithContext` to async because all tests would run at the same time

frankfka: Interesting - any ideas here @comunidadio ?

Another alternative to `import.meta.url === `file://${process.argv[1]}`` is to have an env var where if `true`, the file would run the tests

frankfka: That works. I think we only need a `package.json` script for "full sanity" test suites for each service, as well as a script that tests everything.

`accountSetup` should probably be its own script. We can assume that this is run before any tests are run. Once an acct is setup, we shouldn't need to run `accountSetup` again

Mario857: If we plan to use [built-in test runner](https://tsx.is/node-enhancement#test-runner). I think we would also need to break it down into individual tests. 

```
import { createVertexClient, VertexClient } from '@vertex-protocol/client';
import { prettyPrint } from '../utils/prettyPrint';
import { runWithContext } from '../utils/runWithContext';
import { describe, it } from 'node:test';

describe('[client]: Running query tests', () =>
  void runWithContext((context) => {
    const walletClient = context.getWalletClient();
    const publicClient = context.publicClient;

    const vertexClient: VertexClient = createVertexClient(
      context.env.chainEnv,
      {
        walletClient,
        publicClient,
      },
    );

    const walletClientAddress = walletClient.account.address;

    it('Engine time', async () => {
      prettyPrint(
        'Engine time',
        await vertexClient.context.engineClient.getTime(),
      );
    });

    it('Symbols', async () => {
      prettyPrint(
        'Symbols',
        await vertexClient.context.engineClient.getSymbols({}),
      );
    });
    ....
  }));


```

frankfka:
@comunidadio could you review here as well?

apps/e2e/src/client/orderTests.ts
```diff
@@ -143,13 +117,6 @@ async function orderTests(context: RunContext) {
...
-  await vertexClient.spot.withdraw({
-    subaccountName: 'default',
-    productId: 0,
-    // 1 USDC withdrawal fee
-    amount: initialDepositAmt - addDecimals(1, 6),
-  });
 }
 
-runWithContext(orderTests);
+it('[client]: Running order tests', () => runWithContext(orderTests));
```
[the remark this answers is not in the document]
frankfka: Running account setup before each test would currently balloon the account value over time

Teardown script is interesting, we could get the max withdrawable and withdraw that amount (could also close any positions / cancel orders in the same script)

However, running setup + teardown for every test would make the runtime much longer (b/c of the approval / deposit wait time)

apps/e2e/package.json
```diff
@@ -11,13 +11,26 @@
...
-    "client/collateral-tests": "tsx ./src/client/collateralTests.ts",
-    "client/order-tests": "tsx ./src/client/orderTests.ts",
-    "client/query-tests": "tsx ./src/client/queryTests.ts",
-    "client/ws-message-tests": "tsx ./src/client/wsMessageTests.ts",
-    "engine/full-sanity": "tsx ./src/engine-client/fullSanity.ts",
-    "indexer/full-sanity": "tsx ./src/indexer-client/fullSanity.ts",
-    "trigger/full-sanity": "tsx ./src/trigger-client/fullSanity.ts",
+    "account-setup": "tsx ./src/utils/accountSetup.ts",
+    "client/collateral-tests": "tsx --test ./src/client/collateralTests.ts",
+    "client/full-sanity": "tsx --test ./src/client/**.ts",
```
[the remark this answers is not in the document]
Mario857: Hmm having `"e2e:client": "tsx --test ./src/client` doesn't seem to work.

apps/e2e/README.md
```diff
@@ -3,3 +3,48 @@
...
+
+`yarn account-setup`
+
+## Running Tests
+
+### Client Tests
+
+| Command                      | Description                                                            |
+| ---------------------------- | ---------------------------------------------------------------------- |
+| yarn client/collateral-tests | E2E tests for depositing, withdrawing, and transferring collateral     |
```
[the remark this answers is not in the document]
frankfka: @Mario857 - I think Cedric's suggestion here is to only include the full sanity tests per client. If we want to run individual tests, we can do so via CLI

frankfka:
Will ahve @comunidadio give the final go-ahead here

apps/e2e/README.md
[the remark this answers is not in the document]
Mario857: Okay, I liked the idea of prefixing with e2e. But yes we can drop the individual tests. Lmk.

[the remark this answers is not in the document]
frankfka: Yeah prefixing with e2e works! I think keeping the package.json scripts light is a good idea as well

apps/e2e/src/engine-client/queryTests.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    subaccountOwner: walletClientAddress,
+    subaccountName: 'default',
+  });
+  prettyPrint('Subaccount info', subaccountInfo);
+
+  const symbols = await client.getSymbols({});
+  prettyPrint('Symbols', symbols);
+
+  const products = await client.getAllMarkets();
+  prettyPrint('All products', products);
```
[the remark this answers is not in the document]
Mario857: I like the idea of renaming to `debugPrint` and adding `process.env.DEBUG`.

PR #296, opened 2025-05-09
fix (indexer client types): adjust comments

Mario857:


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -948,17 +948,17 @@ export interface IndexerVrtxSupplySnapshot {
   timestamp: BigDecimal;
   /** VRTX token price in primary quote. */
   vrtxOraclePrice: BigDecimal;
-  /** Total VRTX tokens distributed as staking incentives. */
+  /** Fraction of total VRTX supply distributed as staking incentives.  */
```
frankfka, line 951: So fraction as in - `0.1` for 10%?

I didn't know these were fractions tbh - we should reflect that in the naming - ie. `cumulativeIncentivesFrac/Fraction`

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -253,9 +253,9 @@ export type GetIndexerMultiProductSnapshotsResponse = Record<
 >;
 
 export interface IndexerSnapshotsIntervalParams {
-  // Currently accepts all integers, in seconds
+  /**  Currently accepts all integers, in second */
```
frankfka, line 256: typo here - `seconds`

packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -253,9 +253,9 @@ export type GetIndexerMultiProductSnapshotsResponse = Record<
...
   granularity: number;
-  // Seconds
+  /** In seconds. If maxTimeInclusive is not set, the query may return the earlier timestamps near midnight UTC.*/
```
frankfka, line 258: `the query may return the earlier timestamps near midnight UTC` what does this mean?

Mario857: Would this make more sense
```
  /**
   * Optional upper bound for snapshot timestamps (in seconds).
   * Without this, snapshots may default to align with UTC midnight,
   * which can make "Last 24h" metrics inaccurate.
   */
```

packages/indexer-client/src/types/clientTypes.ts
Mario857: Yes these are fractions per [this](https://docs.google.com/spreadsheets/d/1Z9uKaP6jr9Dsa_e47ci9otZ8fdqOQFHf4_L91vRFvWo/edit?ref=blog.vertexprotocol.com&gid=639807660#gid=639807660) sheet.

kamwithak:
LGTM

PR #297, opened 2025-05-09
implement staking satellite contract

cryptoMavrik:
This pull request introduces support for the `vrtxStakingV2Satellite` contract, enabling staking and reward functionalities on non-canonical chains. Key changes include updates to the `RewardsExecuteAPI` class, the addition of new contract ABI definitions, and modifications to deployment configurations.

### API Enhancements:
* Added new methods to `RewardsExecuteAPI` for staking, withdrawing, and claiming rewards on non-canonical chains, including `stakeV2Satellite`, `withdrawUnstakedV2SatelliteTokens`, and `claimAndStakeSatelliteLiquidTokens`. These methods interact with the `vrtxStakingV2Satellite` contract.
* Updated the `RewardsExecuteAPI` class to include a new parameter type, `StakeV2SatelliteParams`, which supports additional fields like `value` for satellite staking.

### Contract Integration:
* Introduced the `STAKING_V2_SATELLITE_ABI` in `packages/contracts/src/common/abis/StakingV2Satellite.ts`, defining functions such as `stakeAs`, `claimWithdraw`, and `claimAndStake`.
* Registered the `vrtxStakingV2Satellite` contract in `createClientContext` and `validateDeployment` to ensure proper initialization and validation of contract addresses. [[1]](diffhunk://#diff-545914ddb02d3ba335284b78927d49d28e5bbc6c2b7b62723a91c3c716e41bd9R173-R178) [[2]](diffhunk://#diff-f16fd0f499d63d575dbdac4f7e594e051408f8993b9e63edc7db7da83caaf752R154-R156)

### Deployment Updates:
* Added `vrtxStakingV2Satellite` addresses to all deployment configuration files, ensuring compatibility across different environments, including mainnets and testnets. [[1]](diffhunk://#diff-c7b676d27acd6decabd43aef1a45595b5ab443c69316b520cd4d55ee4629bc42L14-R15) [[2]](diffhunk://#diff-8a15b846a19bdf1a3613694174f17f79c02c3ccc00cc6444a364cc9a76ee98f5L14-R15) [[3]](diffhunk://#diff-5da866fb18d3ddceadd77818814cf703161fb5bb26155cd76d213c1d49bd0ce5L14-R15) [[4]](diffhunk://#diff-1a15c05cd6077e491dc62ce533c9fd72e5969b22c808f302684cd70e94504133L14-R15) [[5]](diffhunk://#diff-201191927dc3f135d59022231b39188ed1fa40fe2d39673897e3996ca62c3c81L14-R15) etc.)

### Code Organization:
* Updated `packages/contracts/src/common/abis/index.ts` to export the new `StakingV2Satellite` ABI for broader usability across the codebase.


packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -60,19 +91,32 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
...
     ]);
   }
 
   /**
-   * Unstake VRTX tokens, unstaked tokens that are unlocked will need to be withdrawn
+   * Claim staking rewards (in USDC), swap for VRTX, and stake the VRTX
    */
-  async unstake(params: VrtxTokenAmountParams) {
-    return this.context.contracts.vrtxStaking.write.withdraw([
+  async claimAndStakeStakingRewards() {
```
frankfka, line 105: We could probably do a driveby removal of `claimAndStake` & `stake`

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -93,33 +137,85 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
...
+   * Claim unlocked V2 tokens that were previously unstaked on non-canonical chains
+   */
+  async withdrawUnstakedV2SatelliteTokens(params: CcipParams) {
+    return this.context.contracts.vrtxStakingV2Satellite.write.claimWithdraw({
+      value: toBigInt(params.ccipFee),
+    });
   }
 
+  /**
+   * Claim staking rewards (in USDC) on non-canonical chains
```
frankfka, line 194: Not staking rewards - trading rewards (should probably be grouped with the `claimLiquidTokens` fn)

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -93,33 +137,85 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
...
+   */
+  async claimSatelliteLiquidTokens(params: ClaimLiquidTokensParams) {
+    return this.context.contracts.vrtxStakingV2Satellite.write.claim(
+      await this.getClaimLiquidTokensContractParams(params),
+      { value: toBigInt(params.ccipFee) },
+    );
+  }
+
+  /**
+   * Claim staking rewards (in USDC), swap for VRTX, and stake the VRTX on non-canonical chains
```
frankfka, line 204: There's no swap, trading rewards are in VRTX

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -172,7 +268,14 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
   private async getClaimLiquidTokensContractParams(
     params: ClaimLiquidTokensParams,
   ): Promise<
-    WriteContractParameters<VertexAbis['vrtxAirdrop'], 'claimAndStake'>['args']
+    | WriteContractParameters<
+        VertexAbis['vrtxAirdrop'],
+        'claimAndStake'
+      >['args']
+    | WriteContractParameters<
```
frankfka, line 275: Isn't this the same thing as the fn from `vrtxAirdrop`?

packages/client/src/apis/rewards/types.ts
```diff
@@ -4,12 +4,18 @@ export interface VrtxTokenAmountParams {
...
+  claimAll: true;
+}
+
 // Either specify the amount, or attempt to claim all available tokens
-type AmountOrAllParams =
-  | VrtxTokenAmountParams
-  | {
-      claimAll: true;
-    };
+type AmountOrAllParams = ClaimAllParams | CcipStakingParams;
```
frankfka, line 18: mm but these types apply to contract calls on the canonical chain as well, so why would they include `CcipParams`?

packages/contracts/src/common/deployments/vrtx/deployment.blastSepolia.json
```diff
@@ -11,5 +11,6 @@
   "vrtxLba": "0x0000000000000000000000000000000000000000",
   "vrtxVesting": "0x0000000000000000000000000000000000000000",
   "vrtxStaking": "0x0000000000000000000000000000000000000000",
-  "vrtxStakingV2": "0x0000000000000000000000000000000000000000"
+  "vrtxStakingV2": "0x0000000000000000000000000000000000000000",
+  "vrtxStakingV2Satellite": "0xE71C71C7620ECFD49FaC3A1DE9f6Fcba8b007e57"
```
frankfka, line 15: I thought this was on base?

.gitignore
```diff
@@ -1,4 +1,5 @@
 .nx/
+.eslintcache
```
frankfka, line 2: Wasn't this handled [here](https://github.com/vertex-protocol/vertex-typescript-sdk/pull/299/files)?

cryptoMavrik: I yes I see that now. This was a bad merge.

packages/client/src/apis/rewards/types.ts
cryptoMavrik: Hmm this seems like an oversight on my part.

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
cryptoMavrik: Is `claimAndStake` (both) not configured for v2?

cryptoMavrik: I've removed `stake` but I'm not sure about `claimAndStake`

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -93,33 +143,65 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
...
   }
 
   /**
-   * Claim staking rewards (in USDC), swap for VRTX, and stake the VRTX
+   * Unstake V2 VRTX tokens on non-canonical chains, unstaked tokens that are unlocked will need to be claimed
+   * after 21 day locking period
    */
-  async claimAndStakeStakingRewards() {
-    return this.context.contracts.vrtxStaking.write.claimUsdcAndStake();
+  async unstakeV2SatelliteSlow(params: CcipStakingParams) {
```
frankfka, line 184: Is the thinking that we would have `satellite` as a postfix? If so, this & `withdrawUnstakedV2SatelliteTokens` don't obey that structure

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -93,33 +143,65 @@ export class RewardsExecuteAPI extends BaseVertexAPI {
...
   }
 
   /**
-   * Claim staking rewards (in USDC)
+   * Unstake V2 VRTX tokens on non-canonical chains, unstake tokens instantly with a penalty that is redistributed
+   * to the staking contract
    */
-  async claimStakingRewards() {
-    return this.context.contracts.vrtxStaking.write.claimUsdc();
+  async unstakeV2Satellite(params: CcipStakingParams) {
```
frankfka, line 174: Why does this take in `CcipStakingParams` when there is no need to input amount?

packages/client/src/apis/rewards/types.ts
```diff
@@ -4,13 +4,21 @@ export interface VrtxTokenAmountParams {
   amount: BigDecimalish;
 }
 
+export interface CcipStakingParams extends VrtxTokenAmountParams {
```
frankfka, line 7: nit: function names reference `satellite` but this references `ccip`

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
cryptoMavrik: It is currently a postfix for `v2`. So we would have `v2` & `v2Satellite`. Should we just postfix the entire current method name with "satellite"?

packages/client/src/apis/rewards/types.ts
cryptoMavrik: This was originally referencing "satellite". Would `StakingSatelliteParams` make more sense even though it adds the `ccipFee`?

frankfka: Yes, ccipFee is accurate as it references the fee used for the cross chain tx, but satellite makes sense in the interface naming if all the contracts and fns are named as such

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
cryptoMavrik: Hmm ok that could work.

frankfka:
> I'm still kinda confused by our naming scheme. Is `Satellite` going to be a postfix? If so, why do we have `withdrawUnstakedSatelliteTokens`, etc.?

I don't think this is addressed. Still seeing a mix of naming schemes

cryptoMavrik:
@frankfka this is ready for another look.

packages/client/src/apis/rewards/BaseRewardsAPI.ts
```diff
@@ -0,0 +1,106 @@
...
+      params.args ?? [],
+    );
+  }
+
+  /**
+   * Util function to get estimate of CCIP gas
+   * @param functionIndex
+   * @param encodedAbiParams
+   */
+  async getEstimateCcipGas(functionIndex: number, encodedAbiParams: Address) {
```
frankfka, line 96: This can be local to `RewardsQueryAPI`, right?

`functionIndex` can be typed as well.

`Address` isn't the correct typing for `encodedAbiParams`, are you looking for `Hex`?

packages/client/src/apis/rewards/BaseRewardsAPI.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Util function to encode CCIP abi params
+   * @param params
+   */
+  getEncodedCcipAbiParams(
+    params: Omit<
+      EncodeFunctionDataParameters<VertexAbis['vrtxStakingV2Satellite']>,
+      'abi'
+    >,
+  ) {
+    const functionAbi = this.context.contracts.vrtxStakingV2Satellite.abi.find(
```
frankfka, line 74: Why can't we use `encodeFunctionData`?

frankfka, line 74: Why can't we use `encodeFunctionData`?

frankfka: Not sure we really need this function tbh - individual consumers can call `encodeFunctionData` with the necessary params

packages/client/src/apis/rewards/BaseRewardsAPI.ts
frankfka, line 96: nit: we call this `CcipGas` here, but the execute param is `ccipFee`, we should standardize the naming

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,115 @@
-import { BaseVertexAPI } from '../base';
+import { toBigInt } from '@vertex-protocol/utils';
+import { BaseRewardsAPI } from './BaseRewardsAPI';
+import { ClaimLiquidTokensParams, VrtxTokenAmountParams } from './types';
 
-export class RewardsQueryAPI extends BaseVertexAPI {}
+enum FunctionIndex {
```
frankfka, line 5: nit: this could be more generically named as `SatelliteTransactionType`.

If we don't need a 1-1 mapping of enum name to contract function name, it is usually convention to use `CAPITAL_SNAKE_CASE` (i.e `STAKE_AS`).

It is also [good practice](https://blog.logrocket.com/typescript-enums-vs-types/#typescript-enums-best-practices) to explicitly declare numerical values

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: This is used in `RewardsExecuteAbi` atm.

Good call, I made the enum after the fact.

Hmm possibly. `0x${string}` but hex is likely more of an accurate typing.

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: `encodeFunctionData` doesn't work unfortunately. It doesn't have the same output. I compared the 2 side by side.

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: Hmm yea this aligns with the function name. Maybe we should go with `ccipGas` as well. 🤔

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,124 @@
...
+      '0x',
+    );
+  }
+
+  /**
+   * Get the estimated CCIP fee for the `withdrawSlow` transaction
+   * @returns
+   */
+  async getEstimatedWithdrawSlowCcipFee() {
+    return await this.getEstimateCcipGas(
```
frankfka, line 54: nit: `estimated` or `estimate`? `gas` or `fee`?

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,124 @@
...
+
+  /**
+   * Util function to get estimate of CCIP gas
+   * @private
+   * @param transactionType
+   * @param encodedAbiParams
+   */
+  private async getEstimateCcipGas(
+    transactionType: SatelliteTransactionType,
+    encodedAbiParams: Hex,
```
frankfka, line 113: Possible to default this to `zeroHash`? 

Does it work properly? or are we supposed to just `encodePacked` with `uint8`?

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,124 @@
...
+  }
+
+  /**
+   * Get the estimated CCIP fee for the `claim` transaction
+   * @param params
+   * @returns
+   */
+  async getClaimLiquidTokensCcipFee(params: ClaimLiquidTokensParams) {
+    const encodedAbiParams = encodeAbiParameters(
+      parseAbiParameters('uint32, uint256, uint256, bytes32[]'),
```
frankfka, line 78: This works I think, alternatively, you can see if `viem` exports any utils that can leverage the satellite ABI directly: https://github.com/wevm/viem/blob/main/src/utils/abi/prepareEncodeFunctionData.ts#L85

packages/client/src/apis/rewards/RewardsQueryAPI.ts
cryptoMavrik: Hmm that's an interesting question. For simplicity, it doesn't hurt to have the `encodedAbiParams` even if they are "empty", but the outcome would probably be the same if we just encoded `transactionType`

frankfka: Yes to the first question

packages/client/src/apis/rewards/RewardsQueryAPI.ts
cryptoMavrik: The route view takes is a bit different than our approach. They seem to go down the patch of encoding the function name rather than an index number. Internally they are using after programmatically getting the types from the abi, similarly to the previous iteration, but nothing I see to assist in this step without over-complicating the logic again.

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: To elaborate just a bit, it encodes the function name and then concatenates the two.

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: Switched to `ccipFee` across the board. "Gas" just didn't feel descriptive enough.

packages/client/src/apis/rewards/RewardsQueryAPI.ts
frankfka: Im not suggesting that we replicate their logic, I'm suggesting that we try to use their existing util fns to get the output of `parseAbiParameters` here, whilst using the ABI itself

frankfka:
```tsx
prepareEncodeFunctionData({
  abi: ERC20_ABI,
  functionName: 'approve'
}).abi
```

cryptoMavrik:
```
prepareEncodeFunctionData({
      abi: this.context.contracts.vrtxStakingV2Satellite.abi,
      functionName: 'stakeAs',
    }).abi[0].inputs.map((input) => ({ type: input.type }))
```
This produces the same output as `parseAbiParameters('address, uint128')`

cryptoMavrik: I wasn't implying we replicate it. I was just saying that getting the output `parseAbiParameters('address, uint128')` programmatically from the abi, just complicates the logic. We just need to reduce the inputs to a `{ type: string }[]` which is what  `parseAbiParameters` is doing.

cryptoMavrik:
@frankfka bumping

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,133 @@
...
+  /**
+   * Get the estimated CCIP fee for the `stakeAs` transaction
+   * @param params
+   * @returns
+   */
+  getEstimatedStakeAsCcipFee(params: VrtxTokenAmountParams) {
+    const address = this.getWalletClientAddress();
+
+    if (!address) {
+      throw new Error('No wallet client address found');
```
frankfka, line 27: As a follow up, we may want to extract this to a custom error class (there seems to be a few of this around)

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,133 @@
...
+
+  /**
+   * Util function to get estimate of CCIP Fee
+   * @private
+   * @param transactionType
+   * @param encodedAbiParams
+   */
+  private getEstimatedCcipFee(
+    transactionType: SatelliteTransactionType,
+    encodedAbiParams: Hex = '0x',
```
frankfka, line 107: Not a huge deal, but prior suggestion was to have this default to `zeroHash`

packages/client/src/apis/rewards/RewardsQueryAPI.ts
```diff
@@ -1,3 +1,133 @@
...
+    ]);
+  }
+
+  /**
+   * Util function to get parsed function input data
+   * @private
+   * @param functionName
+   * @returns
+   */
+  private getPrepareEncodeFunctionData(
```
frankfka, line 125: nit: this isn't a simple passthrough for `prepareEncodeFunctionData`, but instead retrieves the abi item & returns the inputs - what about `getFunctionAbiParams` or `getFunctionAbiInputs`?

frankfka (approved):

PR #300, opened 2025-05-15
fix: e2e on esm

comunidadio:
Could not get `ts-node` to work, switching to more modern `tsx` worked out of the box.

Also includes minor import fixes.


packages/contracts/src/common/deployments/index.ts
frankfka, line 1: Whats the context behind this?

frankfka (approved):

frankfka:
> > Perhaps we should add e2e to github actions?
> 
> 
> 
> @frankfka is it worth to follow up with this? We could set private key in GitHub Secrets.

I think its ok for now. E2E currently doesn't fully reset account state, so it can get a bit messy (and probably flakey) if we were to run it on every PR

If we manually trigger the GH action then we can, but the benefit to running it locally might be too marginal?

PR #303, opened 2025-05-27
fix: ensure getTxNonceIfNeeded uses the correct address nonce if...

comunidadio:
…`subaccountOwner` is different from `walletClient.account.address`.

Not sure if we still want to do this after #302, it seems to make sense but is there a use case where the mismatch is possible?

### DIFF
--- a/packages/engine-client/src/EngineExecuteBuilder.ts
+++ b/packages/engine-client/src/EngineExecuteBuilder.ts
@@ -409,12 +409,12 @@ export class EngineExecuteBuilder {
   }
 
   protected async getTxNonceIfNeeded(
-    params: WithBaseEngineExecuteParams<unknown>,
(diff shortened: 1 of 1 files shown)


packages/engine-client/src/EngineExecuteBuilder.ts
```diff
@@ -409,12 +409,12 @@ export class EngineExecuteBuilder {
   }
 
   protected async getTxNonceIfNeeded(
-    params: WithBaseEngineExecuteParams<unknown>,
+    params: WithBaseEngineExecuteParams<unknown> & { subaccountOwner?: string },
```
frankfka, line 412: Is it safe to enforce `{ subaccountOwner: string }`? Could this replace `unknown`?

frankfka (approved):

PR #305, opened 2025-05-30
fix: declare new RewardsQueryAPI functions as async.

comunidadio:
This is for consistency with our other QueryAPIs and to improve error handling so that an exception will always result in a Promise being rejected rather than dangling.


frankfka (approved):
Perhaps my understanding is wrong, but I thought that the fact that the return is a `Promise` would make the async redundant?

PR #310, opened 2025-06-05
integrate claim multiple epoch rewards

cryptoMavrik:
### Refactoring and simplification of rewards APIs:

* [`packages/client/src/apis/rewards/BaseRewardsAPI.ts`](diffhunk://#diff-97fd0a67532196b256704f9daf1b2a91e3dc01953e4af5fbd83db302d7d1164cL1-R46): Refactored `getClaimLiquidTokensContractParams` to remove dependency on `ClaimLiquidTokensParams` and `ClaimLiquidTokensSatelliteParams`. The method now directly retrieves proofs and calculates amounts to claim for all epochs, enabling batch operations.
* [`packages/client/src/apis/rewards/RewardsExecuteAPI.ts`](diffhunk://#diff-f1317adf0779d4e8916bac5a0e2b8d7a4a837ec32756cb0a803431299afaf938L31-R65): Updated methods like `claimLiquidTokens` and `claimAndStakeLiquidTokens` to use batch operations (`claimMultiple` and `claimMultipleAndStake`) instead of single claims. Removed dependency on `ClaimLiquidTokensParams`.
* [`packages/client/src/apis/rewards/RewardsQueryAPI.ts`](diffhunk://#diff-16459fd581627909fd5774b3d92eae003af2a19f5074773231564c02f6e3eddeL71-R72): Modified methods to use batch proofs (`getClaimLiquidTokensCcipFee` and `getClaimAndStakeLiquidTokensCcipFee`), removing the need for individual claim parameters. [[1]](diffhunk://#diff-16459fd581627909fd5774b3d92eae003af2a19f5074773231564c02f6e3eddeL71-R72) [[2]](diffhunk://#diff-16459fd581627909fd5774b3d92eae003af2a19f5074773231564c02f6e3eddeL88-R91)

### Introduction of new types and contract methods:

* [`packages/client/src/apis/rewards/types.ts`](diffhunk://#diff-3c5fd80dbba36c19ce48bf000357b6e7c40e11904349110f27db09adabfc3fbfL13-R29): Added `RewardsLiquidTokensProof` type to represent proofs for batch claims. Removed `ClaimLiquidTokensParams` and `ClaimLiquidTokensSatelliteParams` types, simplifying the API.
* [`packages/contracts/src/common/abis/Airdrop.ts`](diffhunk://#diff-d1ba078c647c6b21acf92f3ddb52f6e55b29fa9f6c02431211e27ddc12430e4bR83-R185): Added new contract methods (`claimMultiple`, `claimMultipleAndStake`) to support batch operations for claiming and staking tokens.

### DIFF
--- a/packages/client/src/apis/rewards/BaseRewardsAPI.ts
+++ b/packages/client/src/apis/rewards/BaseRewardsAPI.ts
@@ -1,56 +1,44 @@
-import { VertexAbis } from '@vertex-protocol/contracts';
-import {
-  toBigDecimal,
-  toBigInt,
-  toIntegerString,
-} from '@vertex-protocol/utils';
-import { WriteContractParameters } from 'viem';
+import { toBigDecimal, toBigInt } from '@vertex-protocol/utils';
 import { BaseVertexAPI } from '../base';
-import {
-  ClaimLiquidTokensParams,
-  ClaimLiquidTokensSatelliteParams,
-} from './types';
+import { RewardsLiquidTokensProof } from './types';
 
 export class BaseRewardsAPI extends BaseVertexAPI {
   /**
    * Util function to share logic between claimLiquidTokens and claimAndStakeLiquidTokens
-   * @param params
+   * @returns Proofs with a claimable amount to claim.
    */
-  async getClaimLiquidTokensContractParams(
-    params: ClaimLiquidTokensParams | ClaimLiquidTokensSatelliteParams,
-  ): Promise<
-    WriteContractParameters<VertexAbis['vrtxAirdrop'], 'claimAndStake'>['args']
-  > {
+  async getClaimAllLiquidTokensContractParams() {
     const address = this.getWalletClientAddress();
 
-    const { totalAmount, proof } = (
-      await this.context.indexerClient.getClaimVrtxMerkleProofs({
-        address,
-      })
-    )[params.epoch];
+    const proofs = await this.context.indexerClient.getClaimVrtxMerkleProofs({
+      address,
+    });
 
-    const airdropContract = this.context.contracts.vrtxAirdrop;
+    const amountsClaimed =
+      await this.context.contracts.vrtxAirdrop.read.getClaimed([address]);
 
-    const amountToClaim = await (async () => {
-      if ('amount' in params) {
-        return params.amount;
-      }
-      const amountsClaimed = await airdropContract.read.getClaimed([address]);
+    return proofs.reduce<RewardsLiquidTokensProof[]>(
+      (acc, { proof, totalAmount }, idx) => {
+        if (idx === 0) return acc; // week 0 is invalid
 
-      const availableAmount = totalAmount.minus(
-        // Some wallets seem to throw a `RangeError` here if we do amountsClaimed[params.epoch]
-        // Likely because `amountsClaimed` isn't a simple array but a proxy
-        toBigDecimal(amountsClaimed.at(params.epoch) ?? 0),
-      );
+        const amountToClaim = totalAmount.minus(
+          // Some wallets seem to throw a `RangeError` here if we do amountsClaimed[params.epoch]
+          // Likely because `amountsClaimed` isn't a simple array but a proxy
+          toBigDecimal(amountsClaimed.at(idx) ?? 0),
+        );
 
-      return toIntegerString(availableAmount);
-    })();
+        if (!amountToClaim.isZero()) {
+          acc.push({
+            epoch: idx,
+            amount: toBigInt(amountToClaim),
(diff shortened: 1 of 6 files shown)


packages/contracts/src/common/abis/Airdrop.ts
frankfka, line 1: Is it me or are a lot of these additions unrelated?

Did we get the ABI from the interface or from the actual contract?

packages/client/src/apis/rewards/RewardsExecuteAPI.ts
```diff
@@ -43,6 +55,18 @@ export class RewardsExecuteAPI extends BaseRewardsAPI {
     );
   }
 
+  /**
+   * Claim earned VRTX tokens and stake them for all epochs that have a claimable amount
+   */
+  async claimMultipleAndStakeLiquidTokens() {
```
frankfka, line 61: This is named `claimMultiple...` but it's actually just claiming all available tokens, which makes the naming a bit confusing

In general, the `claimMultiple...` nomenclature is a bit confusing. ex. `claimMultipleLiquidTokens` would just make me think that I'm claiming >1 tokens, as it makes no reference to epochs.

Also note that `claimMultiple` is just a superset of functionality offered by `claim`, so we can technically replace `claimLiquidTokens` to take in an array of epochs

frankfka: Would help remove duplication from [here](https://github.com/vertex-protocol/vertex-typescript-sdk/pull/310/files#diff-97fd0a67532196b256704f9daf1b2a91e3dc01953e4af5fbd83db302d7d1164cR62) as well

cryptoMavrik: Ok yea I guess my first stab was to use all the available functions on the contract, but you raise some good points. We can likely only use the new functions in the methods already available. I'll refactor all this a bit to clean it up.

packages/client/src/apis/rewards/types.ts
```diff
@@ -10,16 +11,14 @@ export interface SatelliteCcipParams {
...
 
-// Either specify the amount, or attempt to claim all available tokens
-type AmountOrAllParams =
-  | VrtxTokenAmountParams
-  | {
-      claimAll: true;
-    };
-
-export type ClaimLiquidTokensParams = AmountOrAllParams & {
+export interface ClaimProof {
```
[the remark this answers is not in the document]
cryptoMavrik: Yes. To reduce to only using one function, it will probably be best to just have consumers pass in the typed proofs. The sdk still handles the logic, but the params aren't localized to the function anymore.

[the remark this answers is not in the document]
cryptoMavrik: This is bit of a shift from the current implementation, but is cleaner overall. We only should loop through the proofs once, and passing in just an array of epoch numbers would require us to loop again.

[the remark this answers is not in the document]
frankfka: I see. Is there any benefit to this abstraction then? It seems like we're just calling the contract function directly.

Would it work to have the params as something like: 

```tsx
interface ClaimLiquidTokenEpochParams extends VrtxTokenAmountParams {
  epoch: number;
}

type ClaimLiquidTokensParams =
  | {
      claimAll: true;
    }
  | {
      epochs: ClaimLiquidTokenEpochParams[];
    };
```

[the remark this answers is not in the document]
cryptoMavrik: Do we have any use-case for specifying the amount now?

packages/client/src/apis/rewards/BaseRewardsAPI.ts
```diff
@@ -1,56 +1,44 @@
...
-    const airdropContract = this.context.contracts.vrtxAirdrop;
+    const amountsClaimed =
+      await this.context.contracts.vrtxAirdrop.read.getClaimed([address]);
 
-    const amountToClaim = await (async () => {
-      if ('amount' in params) {
-        return params.amount;
-      }
-      const amountsClaimed = await airdropContract.read.getClaimed([address]);
+    return proofs.reduce<RewardsLiquidTokensProof[]>(
```
frankfka, line 20: Could the `RewardsLiquidTokensProof` just be derived from the abi?

ie:
image

frankfka (approved):

packages/client/src/apis/rewards/BaseRewardsAPI.ts
cryptoMavrik: Actually we might be ok if its `readonly`

PR #318, opened 2025-07-16
housekeeping: cleanup chains/staking/rewards

comunidadio:
This PR removes staking/rewards and all chains except Arbitrum from SDK.

Version bump required for building monorepo's #3321


frankfka:
Alright, it seemed easier to rename rather than dig up the removal and reintroduce, but I'm easy

Issue #317, opened 2025-07-06
Error importing the SDK - cannot find module {module_name} imported from "@vertex-protocol/client"


deandreee:
I'm running into issues importing the module into NodeJS + TypeScript project. Been debugging this for days and even created fresh empty project just to replicate this with minimal setup.

I'm using:
```
node: v24
ts-node: "^10.9.2",
typescript: "^5.8.3"
@vertex-protocol/client: "^1.14.0",
```

Basically all I'm doing is creating new NodeJS + TypeScript project, running `npm install @vertex-protocol/client viem bignumber.js` and then just try to import the SDK like

```
import { createVertexClient } from "@vertex-protocol/client";
```

And I get this error:
```
Error: Cannot find module '/home/and/Desktop/code/bera-vertex-esm-fresh/node_modules/@vertex-protocol/client/dist/createVertexClient' imported from /home/and/Desktop/code/bera-vertex-esm-fresh/node_modules/@vertex-protocol/client/dist/index.js
    at finalizeResolution (node:internal/modules/esm/resolve:274:11)
    at moduleResolve (node:internal/modules/esm/resolve:859:10)
    at defaultResolve (node:internal/modules/esm/resolve:983:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:801:12)
    at ModuleLoader.#cachedDefaultResolve (node:internal/modules/esm/loader:725:25)
    at ModuleLoader.#resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:760:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:783:52)
    at ModuleLoader.#cachedResolveSync (node:internal/modules/esm/loader:744:25)
    at ModuleLoader.getModuleJobForRequire (node:internal/modules/esm/loader:462:50)
    at ModuleJobSync.#link (node:internal/modules/esm/module_job:436:34) {
  code: 'ERR_MODULE_NOT_FOUND',
```

I've tried experimenting with switching `type` in `package.json` between `"commonjs"` vs `"module"` and also tried every possible `tsconfig.json` variation (experimenting with `compilerOptions` and `ts-node`) I could think of - result is pretty much the same. 

In commit/version history I see that you've recently migrated the project from CommonJS to ESM module. I think this is causing the issues. 

[...]

PR #320, opened 2025-07-18
fix: safely build ESM+CJS dist for bundler-less environments

comunidadio:
Context:  #317 and https://www.notion.so/Eng-Task-Planning-23059f6be93380ac899fc10859f940b6?p=23259f6be93380c58c7bece7c7bc0642&pm=s

This PR fixes SDK in bundler-less environments (i.e. pure node without webpack/vite/etc).
This makes CI  introduced in #319 complete succesfully as node-compat-test is able to import SDK correctly.

Remaining versioning issue with bogus `viem:*` resolution to be resolved with pnpm migration follow up (using `workspace:` protocol so `lerna` can automatically rewrite published version deps without additional configuration).

### DIFF
--- a/packages/client/tsconfig.build.json
+++ b/packages/client/tsconfig.build.json
@@ -1,7 +1,0 @@
-{
-  "extends": "./tsconfig.json",
-  "compilerOptions": {
-    "composite": false,
-    "incremental": false
-  }
-}
--- a/tsup.base.config.ts
+++ b/tsup.base.config.ts
@@ -8,13 +8,15 @@ export default defineConfig([
         format: ['cjs', 'esm'],
         target: 'esnext',
         outDir: 'dist',
-        dts: true,
+        dts: {
+            compilerOptions: {
+                // Disabling `composite` is required for `tsup` to generate `.d.ts` files correctly.
+                // See https://github.com/egoist/tsup/issues/571
+                composite: false,
+            },
+        },
         sourcemap: true,
         clean: true,
-        // A separate tsconfig to disable `composite` and `incremental` options
-        // which are not compatible with tsup.
-        // See https://github.com/egoist/tsup/issues/647
-        tsconfig: './tsconfig.build.json',
         esbuildPlugins: [
             // This plugin rewrites our extension-less imports to use '.js' as required in bundler-less environments.
             // eslint-disable-next-line @typescript-eslint/no-unsafe-call
--- a/packages/contracts/tsconfig.build.json
+++ b/packages/contracts/tsconfig.build.json
@@ -1,7 +1,0 @@
-{
-  "extends": "./tsconfig.json",
-  "compilerOptions": {
-    "composite": false,
-    "incremental": false
-  }
-}
--- a/packages/engine-client/tsconfig.build.json
+++ b/packages/engine-client/tsconfig.build.json
@@ -1,7 +1,0 @@
-{
-  "extends": "./tsconfig.json",
-  "compilerOptions": {
-    "composite": false,
-    "incremental": false
-  }
-}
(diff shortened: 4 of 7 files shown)


packages/client/src/index.ts
```diff
@@ -13,4 +13,4 @@ export * from '@vertex-protocol/contracts';
 export * from '@vertex-protocol/engine-client';
 export * from '@vertex-protocol/indexer-client';
 export * from '@vertex-protocol/trigger-client';
-export { OptionalSignatureParams } from './apis/types';
+export type { OptionalSignatureParams } from './apis/types';
```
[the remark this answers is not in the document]
frankfka: Do you know if we even need this export?

packages/client/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "composite": false,
```
frankfka, line 4: What is the context behind these changes in `compilerOptions`?

root.tsup.config.ts
```diff
@@ -0,0 +1,27 @@
...
+        sourcemap: true,
+        clean: true,
+        tsconfig: './tsconfig.build.json',
+        esbuildPlugins: [
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+            esbuildPluginFilePathExtensions({ esmExtension: 'js' }),
+            copy({
+                resolveFrom: 'cwd',
+                assets: {
+                    from: ['./src/**/*.json'],
```
frankfka, line 21: Does this retain the path to the `src` folder, or does it dump all `json` files at the root of `dist`?

Does `tsup` not have built-in support for including static files?

root.tsup.config.ts
```diff
@@ -0,0 +1,27 @@
...
+        format: ['cjs', 'esm'],
+        target: 'esnext',
+        outDir: 'dist',
+        dts: true,
+        sourcemap: true,
+        clean: true,
+        tsconfig: './tsconfig.build.json',
+        esbuildPlugins: [
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+            esbuildPluginFilePathExtensions({ esmExtension: 'js' }),
```
frankfka, line 17: What is the purpose of this?

root.tsup.config.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig } from 'tsup';
+import { esbuildPluginFilePathExtensions } from 'esbuild-plugin-file-path-extensions';
+import { copy } from 'esbuild-plugin-copy';
+
+export default defineConfig([
```
frankfka, line 5: I see us adding `sideEffects: false` but not enabling treeshaking here. How come?

frankfka (commented):
Have you tried a local build w/ the frontend?

root.tsup.config.ts
frankfka, line 1: nit: shouldn't the `root` be _after_ `tsup.config` in terms of naming?

Does `root` make sense? or is something like `template` or `base` better?

tsup.base.config.ts
```diff
@@ -0,0 +1,36 @@
...
+                // See https://github.com/egoist/tsup/issues/571
+                composite: false,
+            },
+        },
+        sourcemap: true,
+        clean: true,
+        esbuildPlugins: [
+            // This plugin rewrites our extension-less imports to use '.js' as required in bundler-less environments.
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+            esbuildPluginFilePathExtensions({ esmExtension: 'js' }),
```
frankfka, line 23: Interesting that this is typed as `any`?

frankfka (approved):

PR #312, opened 2025-06-09
housekeeping: migrate to pnpm

comunidadio:
Context: https://www.notion.so/vertexprotocol/Eng-Task-Planning-1bafec14f62e4c1bac82e63aaa8dc68d?p=1ee4976028f080c78763d9a0cb52bd43&pm=s

PR #324, opened 2025-07-23
housekeeping: micro fixes before pnpm migration

comunidadio:
This PR fixes a remaining phantom dependency (@eslint/js) and tweaks scripts to avoid unnecessary `npx` detour.

The phantom dependency fix is required for #312.


frankfka:
> > Have you seen this comment about pnpm? https://eslint.org/docs/latest/use/getting-started#manual-set-up
> 
> 
> 
> Didn't see this. Interesting take by eslint 😆
> 
> That's disabling one of the main benefit of pnpm (no hoisting = no phantom dep) - checking around I do not see any major eslint+pnpm project (nextjs, react-query, react-router, ...) following this advice.
> 
> 

I wonder why they suggested this 🤔

PR #325, opened 2025-08-05
Remove LPs

frankinkfnd:


frankinkfnd:
> Probably should remove `netEntryLpCumulative` and `netEntryLpUnrealized` in `indexer-client/src/dataMappers.ts` and `indexer-client/src/types/clientTypes.ts` ?
> 
> Same probably for a few leftover types in `serverQueryModelTypes.ts`: EngineServerSpotLpBalanceData EngineServerPerpLpBalanceData EngineServerSpotLpState EngineServerPerpLpState
> 
> Not sure it's safe to update now, or the state of LP removal on contracts, we probably want to refresh the ABIs without LP-related methods and fields?

Nice catch!

ABIs will be updated at the very end when backend is done with their re-write

PR #405, opened 2026-04-13
fix e2e: add delays, retry mechanism for clean method

przemekink:


przemekink:
@claude review once

apps/e2e/src/engine-client/builder.test.ts
```diff
@@ -175,14 +175,18 @@ void describe('[engine-client]: builder', () => {
       void test('queries historical order for builder fee', async () => {
         assert.ok(orderDigest, 'orderDigest must be set by previous test');
 
-        await delay(TEST_DELAYS.INDEXER_PROPAGATION);
+        await delay(TEST_DELAYS.INDEXER_PROPAGATION * 10);
```
frankinkfnd: Why is a `* 10` needed? seems a bit weird for us to multiply by magic numbers based on some heuristic

apps/e2e/src/engine-client/executeOps.test.ts
```diff
@@ -280,7 +280,9 @@ void describe('[engine-client]: execute operations', () => {
       debugPrint('Link signer result', linkResult);
       assert.equal(linkResult.status, 'success', 'linkSigner should succeed');
 
-      // Set the linked signer on the client
+      // Let the engine propagate the linked signer state before using it
+      await delay(TEST_DELAYS.BETWEEN_SUITES);
```
frankinkfnd: The semantic meaning `BETWEEN_SUITES` isn't really appropriate here 🤔 

is `SHORT`/`MEDIUM`/`LONG` better for `TEST_DELAYS` then?

apps/e2e/src/engine-client/linkedSigner.test.ts
```diff
@@ -106,6 +106,8 @@ void describe('[engine-client]: linked signer lifecycle', () => {
     assertDefined(result.signer, 'linkedSignerQuery.signer');
 
     tc.engine.setLinkedSigner(linkedSignerWalletClient);
+
+    await delay(TEST_DELAYS.BETWEEN_SUITES);
```
frankinkfnd: Why do we do this at the end of tests in some cases, and at the beginning of tests in others?

apps/e2e/src/engine-client/nlp.test.ts
```diff
@@ -111,17 +111,30 @@ void describe(
...
-      assertDefined(result, 'burnNlpResult');
-      assert.equal(result.status, 'success', 'burnNlp should succeed');
+        debugPrint('Burn NLP result', result);
+        assertDefined(result, 'burnNlpResult');
+        assert.equal(result.status, 'success', 'burnNlp should succeed');
+      } catch (error) {
+        // NLP tokens may still be locked after minting
+        if (
+          error instanceof Error &&
+          (error.message.includes('2096') ||
```
frankinkfnd: Checking the err code should be enough

frankinkfnd: What if we just check the max withdrawable first?

apps/e2e/src/utils/retryWithBackoff.ts
frankinkfnd, line 1: Is it possible to just have the test runner have retries?

frankinkfnd (commented):
Can we figure out a better soln than randomly adding delays (or at the very least we should document the random delays to explain why..?)

apps/e2e/src/engine-client/builder.test.ts
przemekink: Changed to indexer polling, within retries and max timeout. Should be way better than delay's (which might be flaky).

apps/e2e/src/engine-client/executeOps.test.ts
przemekink: Done, changed.

apps/e2e/src/engine-client/linkedSigner.test.ts
przemekink: Improved, now it's consistent

apps/e2e/src/engine-client/nlp.test.ts
przemekink: Changed to withdraw only available balance (if there's any).

apps/e2e/src/utils/retryWithBackoff.ts
przemekink: I don't think so. We can do some custom option

przemekink: There's --test-rerun-failures flag, analyzing

apps/e2e/src/engine-client/linkedSigner.test.ts
```diff
@@ -32,7 +32,7 @@ void describe('[engine-client]: linked signer lifecycle', () => {
   let linkedSignerWalletClient: WalletClientWithAccount;
 
   before(async () => {
-    await delay(TEST_DELAYS.BETWEEN_SUITES * 4);
+    await delay(TEST_DELAYS.LONG * 4);
```
frankinkfnd: Why do we need the `* 4`?

apps/e2e/src/engine-client/nlp.test.ts
```diff
@@ -107,14 +107,26 @@ void describe(
...
+        );
+        return;
+      }
+
+      const burnAmount = BigNumber.min(maxBurnAmount, balanceUnlocked.balance);
       const result = await tc.engine.burnNlp({
         subaccountOwner: tc.walletClientAddress,
         subaccountName: TEST_SUBACCOUNT_NAME,
-        nlpAmount: maxBurnAmount,
+        nlpAmount: burnAmount,
```
frankinkfnd, line 121: For the sake of having tests be re-runnable, should this just burn a very small amt?

apps/e2e/src/utils/retry.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** Maximum number of polling attempts. Defaults to 10. */
+  maxAttempts?: number;
+  /** Delay in milliseconds between polling attempts. Defaults to 1000. */
+  intervalMs?: number;
+}
+
+/**
+ * Polls an async function until the predicate is satisfied.
+ * Used for waiting on eventually-consistent backends (e.g. indexer propagation)
+ * where a fixed delay is unreliable due to variable CI latency.
```
frankinkfnd: Indexer latency should be max 300ms according to Jeury? 🤔

apps/e2e/src/utils/retryWithBackoff.ts
frankinkfnd: What's the update?

przemekink: Image

Seems to work. Testing locally

apps/e2e/src/engine-client/nlp.test.ts
przemekink: Good idea, fixed

apps/e2e/src/utils/retry.ts
przemekink: Improved to match the 300ms

frankinkfnd: Well what i mean is, if the latency isn't that long, shouldn't a static delay work?

przemekink: Ok, once we have the retry mechanism, the simple delay will work (eventually within a flaky test, it will just re-run).

apps/e2e/src/client/collateral.test.ts
```diff
@@ -47,68 +47,95 @@ void describe(
...
+      await delay(TEST_DELAYS.STANDARD);
     });
 
     // ---------------------------------------------------------------
     // Mint, approve, and deposit
     // ---------------------------------------------------------------
     void describe('deposit flow', () => {
-      void test('mints mock ERC20 tokens', async () => {
-        const receipt = await waitForTransaction(
+      let mintReceipt: Awaited<ReturnType<typeof waitForTransaction>>;
```
frankinkfnd: Can you explain these changes?

apps/e2e/src/engine-client/builder.test.ts
```diff
@@ -44,7 +44,7 @@ const INVALID_BUILDER_ID = 999_999;
 
 void describe('[engine-client]: builder', () => {
   before(async () => {
-    await delay(TEST_DELAYS.BETWEEN_SUITES);
+    await delay(TEST_DELAYS.LONG * 2);
```
frankinkfnd, line 48: Can you explain the * 2?

apps/e2e/src/utils/retry.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const DEFAULT_MAX_ATTEMPTS = 3;
+const DEFAULT_INITIAL_DELAY_MS = 500;
+const DEFAULT_MAX_DELAY_MS = 5_000;
+const DEFAULT_BACKOFF_FACTOR = 2;
+
+function isRateLimitError(error: unknown): boolean {
+  if (error instanceof Error) {
+    const msg = error.message.toLowerCase();
+    return msg.includes('429') || msg.includes('too many requests');
```
frankinkfnd: Only need to check code

apps/e2e/src/client/collateral.test.ts
frankinkfnd: Ahh i think we unlocked a rabbit hole here. Looking at your changes I don't think it's smart to rerun only failed tests as our tests are written to depend on each other (even with your changes here)

apps/e2e/src/engine-client/builder.test.ts
przemekink: Yes - using just the `.LONG` caused flakiness. Multiplying the time allowed for more stable execution

apps/e2e/src/client/collateral.test.ts
przemekink: Yeah - that was the initial issue. Then the re-run with changes should make it work. I'll double check

frankinkfnd: What fails without the retry now? We have the execute retry

przemekink: Actually - not to change more files, I'm setting up to re-run the whole test suite, when even one test fails.

frankinkfnd: Isn't that just do-able via rerunning through github or manually?

przemekink: I don't think so. Changed the script to re-run whole test

apps/e2e/package.json
```diff
@@ -17,6 +17,11 @@
     "e2e:indexer": "tsx --test --test-reporter=spec --test-concurrency=1 ./src/indexer-client/*.test.ts",
     "e2e:trigger": "tsx --test --test-reporter=spec --test-concurrency=1 ./src/trigger-client/*.test.ts",
     "e2e:one": "tsx --test --test-reporter=spec --test-concurrency=1",
+    "e2e:retry": "./scripts/run-with-retry.sh ./src/**/*.test.ts",
```
frankinkfnd, line 20: if this is used only in CI then we don't need to add the rest

apps/e2e/src/trigger-client/cancellation.test.ts
```diff
@@ -161,6 +161,20 @@ void describe('[trigger-client]: cancellation', () => {
   });
 
   void describe('post-cancellation queries', () => {
+    before(async () => {
+      // Ensure all trigger orders are cancelled before querying.
```
frankinkfnd: No longer needed right?

frankinkfnd (commented):
Seems like some of the changes added to make individual tests re-run safe can now be reverted

apps/e2e/src/client/order.test.ts
```diff
@@ -181,8 +181,6 @@ void describe('[client]: orders', { timeout: TEST_TIMEOUTS.LONG }, () => {
     });
 
     void test('cancel-and-place replaces the perp order with an IOC order', async () => {
-      assertDefined(perpOrderDigest, 'perpOrderDigest (from prior test)');
```
frankinkfnd, line 184: Why did this fail?

apps/e2e/src/engine-client/builder.test.ts
frankinkfnd: Can you leave that as a comment? 

Is it for rate limits?

apps/e2e/src/engine-client/nlp.test.ts
```diff
@@ -95,26 +92,38 @@ void describe(
...
-      assert.ok(maxBurnAmount.isFinite(), 'maxBurnNlpAmount should be finite');
+      debugPrint('Max burn NLP amount', result);
+      assertDefined(result, 'maxBurnNlpAmount');
+      assert.ok(result.isFinite(), 'maxBurnNlpAmount should be finite');
     });
 
     void test('burnNlp burns available NLP tokens', async () => {
-      if (!maxBurnAmount?.gt(0)) {
-        // No NLP available to burn — nothing to test
+      const { balanceUnlocked } = await tc.engine.getNlpLockedBalances({
```
frankinkfnd: Actually i think max burn amount is more relevant here, because it takes into account subaccount health

apps/e2e/src/engine-client/signerAndOrder.test.ts
```diff
@@ -398,12 +381,10 @@ void describe('[engine-client]: signer and orders', () => {
   // Multi-product order placement and bulk cancellation
   // ---------------------------------------------------------------
   void describe('multi-product order placement and cancellation', () => {
-    before(async () => {
+    void test('places orders for spot and perp products', async () => {
       // Rate-limit delay after the linked signer operations
```
frankinkfnd: Shouldn't this be placed in the `after` clause where the "linked signer operations" are, so we don't depend on test order?

frankinkfnd (commented):
We can also use `retryWithBackoff` in any flaky operations in tests, in case that helps

apps/e2e/src/engine-client/signerAndOrder.test.ts
przemekink: Done

apps/e2e/src/client/order.test.ts
przemekink: reverted back. failed most likely because previous test didn't returned back order digest (overall would be nice to make them in-depended)

frankinkfnd: When would that happen?

i think we can make them independent. just need to include placement and cancel in the same test

przemekink: done, decoupled

apps/e2e/src/client/order.test.ts
```diff
@@ -116,6 +116,11 @@ void describe('[client]: orders', { timeout: TEST_TIMEOUTS.LONG }, () => {
     });
 
     void test('getSubaccountOrders returns orders for the spot product', async () => {
+      await nadoClient.market.placeOrder({
```
frankinkfnd, line 119: Shouldn't we clean this up by canceling?

apps/e2e/src/engine-client/linkedSigner.test.ts
```diff
@@ -32,7 +32,7 @@ void describe('[engine-client]: linked signer lifecycle', () => {
   let linkedSignerWalletClient: WalletClientWithAccount;
 
   before(async () => {
-    await delay(TEST_DELAYS.BETWEEN_SUITES * 4);
+    await delay(TEST_DELAYS.LONG * 3);
```
frankinkfnd, line 36: Comment here as well, please.

apps/e2e/scripts/run-with-retry.sh
frankinkfnd, line 1: Do we still need this after cleaning up the tests and adding retries to some executes?

apps/e2e/src/engine-client/signerAndOrder.test.ts
```diff
@@ -65,16 +65,42 @@ void describe('[engine-client]: signer and orders', () => {
...
   });
 
   // ---------------------------------------------------------------
   // Order placement, querying, and cancellation
   // ---------------------------------------------------------------
   void describe('order placement and queries', () => {
-    let spotOrderDigest: string;
-    let perpIsolatedOrderDigest: string;
-    let marketPrice: { bid: BigNumber; ask: BigNumber };
+    function makeSpotOrder(): EngineOrderParams {
```
frankinkfnd, line 75: Do we need one for perps as well?

apps/e2e/scripts/run-with-retry.sh
przemekink: Yes, I'll monitor the runs, if we won't hit many of re-tries, will get rid of it.

apps/e2e/src/engine-client/linkedSigner.test.ts
przemekink: Done

apps/e2e/src/client/order.test.ts
przemekink: Done

apps/e2e/src/engine-client/signerAndOrder.test.ts
przemekink: Added

apps/e2e/src/client/order.test.ts
```diff
@@ -125,21 +130,22 @@ void describe('[client]: orders', { timeout: TEST_TIMEOUTS.LONG }, () => {
...
       assertDefined(result, 'subaccountOrders');
       assertNonEmptyArray(result.orders, 'subaccountOrders.orders');
+
+      await nadoClient.market.cancelProductOrders({
+        subaccountName: TEST_SUBACCOUNT_NAME,
+        productIds: [TEST_PRODUCT_IDS.SPOT_ETH],
+      });
     });
 
     void test('cancels all spot orders', async () => {
```
frankinkfnd, line 140: Is this now redundant?

apps/e2e/src/engine-client/signerAndOrder.test.ts
```diff
@@ -65,16 +65,68 @@ void describe('[engine-client]: signer and orders', () => {
...
+        amount: EngineOrderParams['amount'];
+        isolated: boolean;
+      }>,
+    ): EngineOrderParams {
+      const amount = opts?.amount ?? addDecimals(-0.01);
+      const appendix = opts?.isolated
+        ? packOrderAppendix({
+            orderExecutionType: 'default',
+            isolated: {
+              margin: addDecimals(shortLimitPrice.multipliedBy(0.03).div(10)),
```
frankinkfnd: Ahh maybe this doesn't work in general. This implies 10x leverage

apps/e2e/src/trigger-client/listingQueries.test.ts
```diff
@@ -62,7 +63,7 @@ void describe(
           type: 'price',
           criteria: {
             type: 'mid_price_above',
-            triggerPrice: midPrice.multipliedBy(1.5),
+            triggerPrice: removeDecimals(midPrice.multipliedBy(1.5)),
```
frankinkfnd: Was this intended? Not sure removing decimals makes sense here? same w/ L114

apps/e2e/src/trigger-client/placement.test.ts
```diff
@@ -126,7 +127,7 @@ void describe('[trigger-client]: placement', () => {
         type: 'price',
         criteria: {
           type: 'oracle_price_below',
-          triggerPrice: toBigNumber(60000),
+          triggerPrice: removeDecimals(toBigNumber(60000)),
```
frankinkfnd: huh..?

apps/e2e/src/engine-client/signerAndOrder.test.ts
frankinkfnd: Bumping this, not sure `makePerpOrder` is gonna be useful / scalable

apps/e2e/src/utils/retryWithBackoff.ts
frankinkfnd, line 1: nit: `retryWithBackoff.ts`

frankinkfnd (approved):

PR #409, opened 2026-05-06
fix: e2e

przemekink:


apps/e2e/src/utils/helpers.ts
```diff
@@ -0,0 +1,15 @@
+import { removeDecimals } from '@nadohq/shared';
+import BigNumber from 'bignumber.js';
+
+/** Whole tick (price_increment_x18 = 10^18): limit prices must land on integer units after x18 scaling */
+export function alignPriceTick(price: BigNumber): BigNumber {
```
frankinkfnd: Is this applied on x18 numbers? if so we have `toIntegerString`. If not then we should round to price increment

apps/e2e/src/trigger-client/placement.test.ts
```diff
@@ -42,7 +44,7 @@ void describe('[trigger-client]: placement', () => {
     const marketPrice = await tc.engine.getMarketPrice({
       productId: TEST_PRODUCT_IDS.SPOT_ETH,
     });
-    midPrice = removeDecimals(marketPrice.ask.plus(marketPrice.bid).div(2));
+    midPrice = marketPrice.ask.plus(marketPrice.bid).div(2);
```
frankinkfnd, line 44: Why not just `removeDecimals` here once?

apps/e2e/src/indexer-client/leaderboardTests.test.ts
```diff
@@ -77,7 +77,11 @@ function assertLeaderboardContestShape(
   assertBoolean(contest.active, `${label}.active`);
   assertBigNumberFinite(contest.lastUpdated, `${label}.lastUpdated`);
   assertString(contest.title, `${label}.title`);
-  assertString(contest.description, `${label}.description`);
+  assert.equal(
```
frankinkfnd, line 80: Isn't this doing the same thing as `assertString`?

apps/e2e/src/utils/helpers.ts
przemekink: Replaced

apps/e2e/src/trigger-client/placement.test.ts
przemekink: Replaced

apps/e2e/src/indexer-client/leaderboardTests.test.ts
przemekink: `assertString` requires non-empty string. The equal allows for empty string.

frankinkfnd: Maybe `assertString(val, label, allowEmpty)` or `assertNonEmptyString`?

Currently this is not clear

apps/e2e/src/utils/helpers.ts
```diff
@@ -0,0 +1,11 @@
+import { toBigNumber, toIntegerString } from '@nadohq/shared';
+import BigNumber from 'bignumber.js';
+
+/**
+ * Rounds a human-readable price to the nearest whole unit. Required because
+ * price_increment_x18 = 10^18 — after x18 scaling, prices must land on integer
+ * multiples of 10^18, i.e. whole-unit increments in human terms.
+ */
+export function alignPriceTick(price: BigNumber): BigNumber {
```
frankinkfnd: Oh hmm I actually think you can just call `dp(0)` on the `price` itself

frankinkfnd: We can keep this if you'd like as `toIntegerBigNumber`, or just have `dp(0)` everywhere

frankinkfnd (approved):

PR #410, opened 2026-05-07
add country code query

bertinkfnd:
This pull request adds support for retrieving the user's country code as detected by the gateway, enabling region-based UI gating such as geoblocking. The main changes include adding a new method to the `EngineWebClient` for fetching the country code, updating the type definitions, and introducing corresponding tests.

**New country code detection support:**

* Added a new method `getCountryCode` to `EngineWebClient` that retrieves the ISO 3166-1 alpha-2 country code from the `x-nado-country` response header on the `/ip` endpoint. Returns `null` if not available.
* Introduced the `GetEngineCountryCodeResponse` type (`string | null`) in `clientQueryTypes.ts` to represent the country code or absence thereof.
* Added a constant `COUNTRY_HEADER` for the response header key in `EngineWebClient.ts`.

**Testing:**

* Added a test in `query.test.ts` to verify that `getCountryCode` returns a valid ISO country code or `null`.

**Other improvements:**

* Fixed a comment in `getIpBlockStatus` to clarify that the `/ip` endpoint is used (not `/time`).

### DIFF
--- a/packages/engine-client/src/EngineWebClient.ts
+++ b/packages/engine-client/src/EngineWebClient.ts
@@ -1,9 +1,12 @@
 import { EngineBaseClient } from './EngineBaseClient';
 import {
   EngineServerIpBlockResponse,
+  GetEngineCountryCodeResponse,
   GetEngineIpBlockStatusResponse,
   GetEngineTimeResponse,
 } from './types';
+
+const COUNTRY_HEADER = 'x-nado-country';
 
 /**
  * Queries that talk directly to web, _not_ the engine. Placing here in the `engine-client` as we don't have enough
@@ -16,7 +19,7 @@ export class EngineWebClient extends EngineBaseClient {
   async getIpBlockStatus(): Promise<GetEngineIpBlockStatusResponse> {
     return (
       this.axiosInstance
-        // Use the /time endpoint and listen to 403 responses
+        // Use the /ip endpoint and listen to 403 responses
         .get(`${this.opts.url}/ip`, {
           // IP checks go through Cloudflare, which uses allow-origin as *, so withCredentials needs to be false
           withCredentials: false,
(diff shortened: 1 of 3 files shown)


packages/engine-client/src/EngineWebClient.ts
```diff
@@ -36,6 +39,23 @@ export class EngineWebClient extends EngineBaseClient {
...
+   * Useful for region-based UI gating such as geoblocking trading competitions.
+   */
+  async getCountryCode(): Promise<GetEngineCountryCodeResponse> {
+    const res = await this.axiosInstance.get(`${this.opts.url}/ip`, {
+      // IP checks go through Cloudflare, which uses allow-origin as *, so withCredentials needs to be false
+      withCredentials: false,
+    });
+
+    const rawCountry: unknown = res.headers?.[COUNTRY_HEADER];
+    return typeof rawCountry === 'string' ? rawCountry : null;
```
frankinkfnd: What other type would it be?

packages/engine-client/src/EngineWebClient.ts
```diff
@@ -1,10 +1,13 @@
 import { EngineBaseClient } from './EngineBaseClient';
 import {
   EngineServerIpBlockResponse,
+  GetEngineCountryCodeResponse,
   GetEngineIpBlockStatusResponse,
   GetEngineTimeResponse,
 } from './types';
 
+const COUNTRY_HEADER = 'x-nado-country';
```
frankinkfnd, line 9: Does casing not matter here?

bertinkfnd: afaik it needs to be lower-case for axios.

bertinkfnd: *Axios stores response header keys in lowercase. Whether the underlying transport is Node's http module (which lowercases via IncomingMessage.headers) or the browser's XMLHttpRequest/fetch, axios parses the header block and stores keys lowercased on res.headers. Bracket-indexing is case-sensitive on plain JS objects, so the key you index with must be lowercase.*

frankinkfnd (approved):

PR #412, opened 2026-05-23
housekeeping: upgrade dependencies

cedricinkfnd:
Context: https://linear.app/inkfnd-nado/issue/ENGW-171/upgrade-websdk-deps

This PR upgrade dependencies.

Noteworthy upgrades:
- Upgrade to TypeScript 6.0 (so we're ready for Go-based 7.0) - some of the other updates are for TS 6.0 support, including [now resolved ts-jest blocker](https://github.com/nadohq/nado-typescript-sdk/pull/412/changes#r3292359517).
- Migrated to tsdown as tsup is EOL.

As Viem is also upgraded, should sync merge with the web side.


bunfig.toml
```diff
@@ -2,4 +2,4 @@
 # Add packages to minimumReleaseAgeExcludes when a newer version is required.
 [install]
 minimumReleaseAge = 259200  # 3 days, in seconds
-minimumReleaseAgeExcludes = []
```
cedricinkfnd, line 5: we can remove this before merge.

29.4.11 was released 2 days ago which includes https://github.com/kulshekhar/ts-jest/commit/39418187515f11b6584d35a4e3ddf50231f74936 that fixed an annoying blocker for TS 6.0 with our SDK when building CJS (ESM was fine).

packages/trigger-client/package.json
```diff
@@ -43,16 +43,16 @@
...
     "@nadohq/shared": "^0.15.0",
-    "axios": "1.14.0",
+    "axios": "1.16.1",
     "ts-mixer": "6.0.4"
   },
   "peerDependencies": {
-    "bignumber.js": "10.0.2",
-    "viem": "2.45.3"
+    "bignumber.js": "11.1.1",
+    "viem": "2.49.3"
```
frankinkfnd, line 51: When we bump SDK we need to bump this as well on web

frankinkfnd: Same with `bignumber.js`

frankinkfnd (commented):
For `bignumber.js`, our `toBigInt` can likely use the newly added function internally: https://github.com/MikeMcl/bignumber.js/releases/tag/v11.1.0

cedricinkfnd:
> For `bignumber.js`, our `toBigInt` can likely use the newly added function internally: https://github.com/MikeMcl/bignumber.js/releases/tag/v11.1.0

Checked but it's not really practical in our case, the function returns `bigint | null` so we would need a guard to throw on NaNs/inf (instead of letting the current BigInt constructor throw) and we would also need to call the `toFixed(0)` from `toIntegerString` anyway.

bunfig.toml
cedricinkfnd: Removed it from `minimumReleaseAgeExcludes` as we're past the 3 days window now.

frankinkfnd (approved):

cedricinkfnd:
@frankinkfnd PR #4229 on web side cleared remaining "unknown". Merging. Please bump version when you have some time.

frankinkfnd:
@cedricinkfnd `0.17.0`

PR #417, opened 2026-06-09
update trading comp types

bertinkfnd:
This pull request updates the leaderboard qualification status types in both client and server model definitions to support more detailed qualification reasons. This allows the system to represent additional cases where a participant may not qualify, improving clarity and consistency between client and server.

**Leaderboard qualification status enhancements:**

* Expanded the `qualificationStatus` field in `IndexerLeaderboardTrackPosition` (client type) to include `'insufficient_volume'` and `'insufficient_account_value_and_volume'`, in addition to the existing statuses.
* Expanded the `qualification_status` field in `IndexerServerLeaderboardTrackPosition` (server type) with the same new statuses to keep client and server models aligned.

### DIFF
--- a/packages/indexer-client/src/types/clientTypes.ts
+++ b/packages/indexer-client/src/types/clientTypes.ts
@@ -581,7 +581,11 @@ export interface IndexerSocialAccountInfo {
 export interface IndexerLeaderboardTrackPosition {
   value: BigNumber;
   rank: BigNumber;
-  qualificationStatus: 'qualified' | 'insufficient_account_value';
+  qualificationStatus:
+    | 'qualified'
+    | 'insufficient_account_value'
+    | 'insufficient_volume'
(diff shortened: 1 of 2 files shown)


packages/indexer-client/src/types/clientTypes.ts
```diff
@@ -581,7 +581,11 @@ export interface IndexerSocialAccountInfo {
 export interface IndexerLeaderboardTrackPosition {
   value: BigNumber;
   rank: BigNumber;
-  qualificationStatus: 'qualified' | 'insufficient_account_value';
+  qualificationStatus:
+    | 'qualified'
+    | 'insufficient_account_value'
```
frankinkfnd, line 586: Do we really need this as a union type? why do we not just have `isQualified` as a boolean? And perhaps return tracked volume / account value to compare if really needed (which we don't need rn?)

PR #418, opened 2026-06-11
Add edge cached queries

frankinkfnd:
Based on: https://docs.nado.xyz/developer-resources/api/gateway/edge

### DIFF
--- a/packages/engine-client/src/EngineEdgeQueryClient.ts
+++ b/packages/engine-client/src/EngineEdgeQueryClient.ts
@@ -1,0 +1,211 @@
+import { mapValues, MarketWithProduct } from '@nadohq/shared';
+import { EngineBaseClient } from './EngineBaseClient';
+import {
+  EngineCachedTimeParams,
+  EnginePingParams,
+  EnginePingResponse,
+  EngineServerStatusResponse,
+  EngineSymbolsResponse,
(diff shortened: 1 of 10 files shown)


frankinkfnd:
@claude review once

packages/engine-client/src/EngineEdgeQueryClient.ts
```diff
@@ -0,0 +1,211 @@
...
+  async ping(params: EnginePingParams = {}): Promise<EnginePingResponse> {
+    const baseResponse = await this.edgeControlQuery('ping', {
+      id: params.id,
+      client_time:
+        params.clientTime != null ? String(params.clientTime) : undefined,
+    });
+    return {
+      id: baseResponse.id,
+      serverTime: Number(baseResponse.server_time),
+      clientTime:
```
bertinkfnd, line 191: Could this be different than what would be passed into `params`?

bertinkfnd (approved):

packages/engine-client/src/EngineEdgeQueryClient.ts
frankinkfnd: I don't think so

frankinkfnd:
@bertinkfnd this will be published

PR #430, opened 2026-07-21
feat: add support for update_dependency in trigger-client

cedricinkfnd:
Context: https://docs.nado.xyz/developer-resources/api/trigger/executes/update-dependency

### DIFF
--- a/apps/e2e/src/trigger-client/updateDependency.test.ts
+++ b/apps/e2e/src/trigger-client/updateDependency.test.ts
@@ -1,0 +1,221 @@
+import {
+  EngineOrderParams,
+  EnginePlaceOrderParams,
+} from '@nadohq/engine-client';
+import {
+  addDecimals,
+  BigNumberish,
+  getOrderNonce,
+  getOrderVerifyingAddress,
+  packOrderAppendix,
+} from '@nadohq/shared';
+import { TriggerPlaceOrderParams } from '@nadohq/trigger-client';
+import assert from 'node:assert/strict';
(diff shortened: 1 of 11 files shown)


packages/engine-client/src/types/serverExecuteTypes.ts
```diff
@@ -36,6 +37,7 @@ export interface EngineServerExecuteResponseDataByType {
   place_order: EngineServerPlaceOrderResponse;
   place_orders: EngineServerPlaceOrdersResponse;
   transfer_quote: null;
+  update_dependency: { digest: string };
```
frankinkfnd: Nit: can we create something like `EngineServerPlaceOrderResponse`?

apps/e2e/src/trigger-client/updateDependency.test.ts
frankinkfnd, line 1: Shouldn't the tests be using the digest of a limit order as the dependency of a trigger order?

I am not sure using a trigger order's digest as a dependency is actually supported..

cedricinkfnd: Oops right, interestingly the current test passed so it is supported (?).

Updated.

packages/engine-client/src/types/serverExecuteTypes.ts
cedricinkfnd: done.

frankinkfnd (approved):

frankinkfnd:
`0.26.0` is published

PR #432, opened 2026-07-23
housekeeping: map backend error codes

cedricinkfnd:
Context: https://linear.app/inkfnd-nado/issue/ENGW-219/map-all-error-codes-into-the-sdk


apps/e2e/src/errorCodes.test.ts
frankinkfnd, line 1: Seems more like a unit test though i suppose we wouldn't have a good place to put it

frankinkfnd (approved):

cedricinkfnd:
@frankinkfnd can you please bump

